PR #2, opened 2019-06-05
build: update package-lock.json and .gitignore

albemarle:


package.json
```diff
@@ -108,7 +108,7 @@
     "purgecss-webpack-plugin": "^1.5.0",
     "react-dev-utils": "^5.0.0",
     "react-test-renderer": "^16.8.6",
-    "reactifex": "1.1.1",
+    "reactifex": "^1.1.1",
```
robrap: Do we have a ticket to get off reactifex?  We need one.  Can the code just move to the new i18n repo?

albemarle: We used to have ARCH-518 but we closed it, I believe because we were going to include it with https://openedx.atlassian.net/browse/ARCH-654 .  Worth reopening?

PR #4, opened 2019-06-06
populate page with mock data

abutterworth:
Screen Shot 2019-06-06 at 2 23 58 PM

### DIFF
--- a/src/payment/PaymentPage.jsx
+++ b/src/payment/PaymentPage.jsx
@@ -64,7 +64,7 @@ class PaymentPage extends React.Component {
       orderTotal,
     } = this.props;
     const loaded = !loading && !loadingError;
-    const basketHasItems = orderTotal;
+    const basketHasItems = orderTotal !== undefined;
 
     return (
       <div className="page__payment container-fluid py-5">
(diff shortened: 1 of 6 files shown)


src/payment/PaymentPage.jsx
```diff
@@ -40,29 +40,48 @@ class PaymentPage extends React.Component {
...
   render() {
     const {
       loading,
       loadingError,
-      payment,
+      orderTotal,
     } = this.props;
     const loaded = !loading && !loadingError;
-    const hasPayments = payment.total;
+    const basketHasItems = orderTotal !== undefined;
```
douglashall: I wonder if the API response should be more explicit instead of putting this business logic here, i.e. return a `basketHasItems` value in the response.

abutterworth: We could also use the length of `products`

douglashall: Exactly my point. That's a decision better made by the backend.

douglashall: Something to consider down the road...

PR #6, opened 2019-06-07
Set codecov’s target to 100%.

davidjoy:
Let's see if this works.


.travis.yml
```diff
@@ -18,6 +18,6 @@ script:
   - npm run test
   - npm run build
   - npm run is-es5
+  - codecov
```
davidjoy: I moved this here hoping that if it DID have a non-zero exit code, Travis would bail on it.  No such luck.

PR #18, opened 2019-06-12
Implement card holder information part of the payment form.

douglashall:


src/payment/PaymentForm.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { reduxForm } from 'redux-form';
+import { FormattedMessage } from '@edx/frontend-i18n';
+
+import CardHolderInformation from './CardHolderInformation';
+
+class PaymentForm extends React.Component {
+  onSubmit(values) {
+    /* istanbul ignore next */
```
abutterworth: istanbul?

douglashall: Found this: https://disjoint.ca/til/2016/10/30/excluding-code-blocks-from-jest-coverage-reports/

This is temporary until we actually implement this function for realz.

src/payment/PaymentForm.jsx
```diff
@@ -0,0 +1,35 @@
...
+      </form>
+    );
+  }
+}
+
+PaymentForm.propTypes = {
+  handleSubmit: PropTypes.func.isRequired,
+};
+
+export default reduxForm({ form: 'payment' })(PaymentForm);
```
abutterworth: This key `form` needs to match the key in `combineReducers` right? Might be worth making a note or exporting a constant like `storeName` as it's done for `paymentStoreName`

src/payment/CardHolderInformation.jsx
```diff
@@ -0,0 +1,149 @@
...
+import { Field } from 'redux-form';
+import { injectIntl, intlShape, FormattedMessage } from '@edx/frontend-i18n';
+
+import { countryOptionsSelector } from './data/selectors';
+import messages from './CardHolderInformation.messages';
+
+class CardHolderInformation extends React.Component {
+  constructor(props) {
+    super(props);
+    this.countryOptions = [{
```
davidjoy: One thing to note is that by doing this in the constructor, it'll only happen once when the CardHolderInformation component is instantiated, prior to any lifecycle methods.  If the countryOptions prop is updated/gets touched for whatever reason, it'll lose this added value.  Doing this in renderCountryOptions would ensure it always happens, and would remove the need for the instance variable.

src/payment/CardHolderInformation.jsx
```diff
@@ -0,0 +1,149 @@
...
+            id="payment.card.holder.information.heading"
+            defaultMessage="Card Holder Information"
+            description="The heading for the credit card holder information form"
+          />
+        </h2>
+
+        <div className="row">
+          <div className="col-lg-6 form-group">
+            <label htmlFor="firstName">
+              <FormattedMessage
```
davidjoy: I believe using <FormattedMessage> will cause it to insert an unncecessary <span> element in here.  (I _think_ - worth double checking me).  We might want to use `this.props.intl.formatMessage()` instead to avoid the unnecessary DOM element.

src/payment/CardHolderInformation.jsx
douglashall: Sounds good. This was a copy/paste from one of the other apps.

src/payment/CardHolderInformation.jsx
douglashall: Is there a negative consequence of the unnecessary DOM element?

PR #19, opened 2019-06-13
Coupon UI

davidjoy:


package.json
```diff
@@ -141,13 +141,17 @@
...
+      "src/common/serviceUtils.js",
       "src/common/index.js",
       "src/components",
       "src/i18n",
       "src/store",
       "environment.js",
       "src/data",
-      "src/payment/data"
+      "src/payment/data",
+      "src/payment/coupon/data/service",
```
davidjoy: Tests for these two files (service and sagas) are forthcoming - we decided to make this PR now and get it in so we can keep moving.

src/common/serviceUtils.js
```diff
@@ -0,0 +1,59 @@
+import pick from 'lodash.pick';
```
davidjoy: This file is from frontend-app-account and should ultimately pulled out into the frontend-common repo

src/payment/__mocks__/defaultState.mockStore.js
```diff
@@ -24,8 +24,18 @@ module.exports = {
     levelOfEducation: null
   },
   payment: {
-    loading: false,
-    loadingError: null,
+    basket: {
```
davidjoy: This redux store organization is what I was talking about in slack the other day - more modular/granular.  You'll see it come up elsewhere in this PR.

src/payment/coupon/Benefit.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { injectIntl, intlShape } from '@edx/frontend-i18n';
+
+import messages from './messages';
+import { PERCENTAGE_BENEFIT, ABSOLUTE_BENEFIT } from './data/constants';
+
+const Benefit = (props) => {
```
davidjoy: This component displays a description of the applied coupon benefit.

src/payment/coupon/Benefit.jsx
```diff
@@ -0,0 +1,54 @@
...
+        style: 'currency',
+        currency: 'USD',
+        minimumFractionDigits: 0,
+      });
+      formattedMessage = intl.formatMessage(messages['payment.coupon.benefit.absolute'], {
+        code,
+        value: formattedValue,
+      });
+      break;
+    default:
```
davidjoy: This default case seemed valuable to me; the alternative is showing nothing, which seems particularly bad fallback behavior.

src/payment/coupon/data/reducers.js
```diff
@@ -0,0 +1,68 @@
+import { ADD_COUPON, REMOVE_COUPON, UPDATE_COUPON_DRAFT } from './actions';
+
+const defaultState = {
+  benefit: null,
```
davidjoy: `benefit` describes what the coupon does to the price.  `code` is something like "SUMMER20".  `voucherId` is a numeric ID that describes the particular, applied coupon instance.

src/payment/data/reducers.js
```diff
@@ -1,11 +1,14 @@
...
 
-export const initialState = {
-  loading: true,
+export const basketInitialState = {
+  loading: false,
   loadingError: null,
 };
 
-const reducer = (state = initialState, action) => {
+const basket = (state = basketInitialState, action = null) => {
```
davidjoy: This `basket` reducer is a catch-all for all the data that comes back that hasn't yet been split out into its own sub-modules, such as `summary` or `products`, perhaps.  Point being, I wanted to move it out of the way of `coupon` start using combineReducers below, but haven't yet messed with what it actually does.

src/payment/data/sagas.js
```diff
@@ -10,14 +10,19 @@ import {
 // Services
 import * as PaymentApiService from './service';
 
+import { saga as couponSaga, addCouponSuccess, addCouponBegin } from '../coupon';
 
 export function* handleFetchBasket() {
```
davidjoy: handleFetchBasket handles the request to get the complete basket state and then the downstream modifications to the redux store.  Therefore, it needs to call into the coupon actions to update the coupon state.

src/payment/PaymentPage.jsx
```diff
@@ -73,6 +77,7 @@ class PaymentPage extends React.Component {
         {loaded ? (
           <div className="row">
             <div className="col-6">
+              <CouponForm />
```
abutterworth: Doesn't the coupon form show up inside the BasketSummary?

davidjoy: I haven't yet put it in the right place, admittedly.  It's on my coupon todo list. :)

src/payment/coupon/data/redux.test.js
```diff
@@ -0,0 +1,93 @@
...
+    });
+  });
+
+  it('should match default state', () => {
+    expect(store.getState()).toMatchSnapshot();
+  });
+
+  it('should begin to add a coupon', () => {
+    store.dispatch(addCouponBegin());
+    expect(store.getState()).toMatchSnapshot();
```
abutterworth: Cool beans!

src/payment/coupon/Benefit.jsx
albemarle: Would that be a useful header comment?

albemarle: In general the component names have been pretty self-explanatory, but I think Benefit is less so.

src/payment/coupon/Benefit.jsx
```diff
@@ -0,0 +1,54 @@
...
+      // Creates a generic message for sanity, as it's possible the server might send up a type we
+      // don't know about and it'd be better to display this message than to show nothing.
```
albemarle: +1 on the comment, though I think it can be shorter.

`// In case we didn't recognize the type` or something along those lines.

src/payment/coupon/CouponForm.jsx
```diff
@@ -0,0 +1,120 @@
...
+        </ValidationFormGroup>
+      </form>
+    );
+  }
+
+  renderRemove() {
+    const { intl } = this.props;
+    return (
+      <form onSubmit={this.handleRemoveSubmit} className="d-flex align-items-center mb-3">
+        {this.props.benefit !== null ? (
```
albemarle: When there's both an if and an else part, it's often easier to read if the condition lacks ! .

src/payment/coupon/data/reducers.js
albemarle: Would these be useful comments to add to the code?

PR #20, opened 2019-06-14
Handling known server errors for coupons.

davidjoy:
Note that this will be more testable once we get actual responses from the server.  For now the unit tests are the way in which I proved it worked.


src/payment/coupon/CouponForm.jsx
```diff
@@ -33,16 +33,69 @@ export class CouponForm extends Component {
     this.props.removeCoupon(this.props.voucherId);
   }
 
+  renderInvalidMessage() {
```
davidjoy: There may be a more concise way to do this; I didn't want to presume they'd all need the code (even though all the current ones do), otherwise I could have collapsed this down a lot.

davidjoy: Just letting it breathe until the code speaks to me and tells me what shape it wants to be. ;)

src/common/serviceUtils.js
```diff
@@ -41,7 +41,9 @@ export function unpackFieldErrors(fieldErrors) {
  * Processes and re-throws request errors.  If the response contains a field_errors field, will
  * massage the data into a form expected by the client.
  *
- * Field errors will be packaged as an api error with a fieldErrors field usable by the client.
+ * If the response contains a single APi error, will similarly format that for the client.
```
douglashall: `API`

src/payment/coupon/CouponForm.jsx
```diff
@@ -33,16 +33,69 @@ export class CouponForm extends Component {
...
+          code,
+        });
+        break;
+      case 'code_not_valid':
+        invalidMessage = intl.formatMessage(messages['payment.coupon.error.code_not_valid'], {
+          code,
+        });
+        break;
+      default:
+        invalidMessage = intl.formatMessage(messages['payment.coupon.error.unknown']);
```
albemarle: Should we log here?

src/common/serviceUtils.js
davidjoy: 😱😱😱

src/payment/coupon/CouponForm.jsx
davidjoy: Good point!  Yeah, maybe... let me add something.

src/payment/coupon/CouponForm.jsx
```diff
@@ -33,16 +33,69 @@ export class CouponForm extends Component {
...
 
+  renderInvalidMessage() {
+    const { code, errorCode, intl } = this.props;
+    if (errorCode === null) {
+      return null;
+    }
+    let invalidMessage = null;
+    switch (errorCode) {
+      case 'empty_basket':
+        invalidMessage = intl.formatMessage(messages['payment.coupon.error.empty_basket'], {
```
albemarle: I'd probably either make a map from error code to message id, or do a shorter version of the case:

`
case 'empty_basket':
case 'too_many_kittens':
case 'missing_cute_bow':
    messageId = 'payment.coupon.error.' + errorCode; // not going to try to mess with Markdown and backticks in this comment!
    break;
default:
    messageId = 'payment.coupon.error.doh';
}

return intl.formatMessage(messages[messageId, { code });
`

PR #22, opened 2019-06-14
Implement credit card details form.

douglashall:


src/payment/CardDetails.jsx
```diff
@@ -0,0 +1,95 @@
...
+        <div className="row">
+          <div className="col-lg-6 form-group">
+            <label htmlFor="cardNumber">
+              <FormattedMessage
+                id="payment.card.details.number.label"
+                defaultMessage="Card Number (required)"
+                description="The label for the required credit card number field"
+              />
+            </label>
+            <Field name="cardNumber" component="input" type="text" required className="form-control" />
```
davidjoy: I was just noticing after adding react-axe - the fields need to have an `id` attribute for the labels' `htmlFor` to hook up to.  They don't actually point at `name` (which is what I thought for a while)

davidjoy: Both card number and security code are supposed to be secure/password fields, I think.

PR #24, opened 2019-06-14
Saga and service tests.

davidjoy:


src/payment/coupon/data/sagas.js
```diff
@@ -26,14 +26,14 @@ function* handleAddCoupon(action) {
   }
 }
 
-function* handleRemoveCoupon(action) {
+export function* handleRemoveCoupon(action) {
   yield put(removeCouponBegin());
   try {
-    const result = yield call(postCoupon, action.payload.codel);
+    const result = yield call(deleteCoupon, action.payload.voucherId);
```
davidjoy: Straight up bug.

src/payment/coupon/data/sagas.js
```diff
@@ -26,14 +26,14 @@ function* handleAddCoupon(action) {
...
     yield put(removeCouponSuccess(result));
   } catch (e) {
     if (e.code !== undefined) {
-      removeCouponFailure(e.code);
+      yield put(removeCouponFailure(e.code));
```
davidjoy: Another straight up bug.

src/payment/coupon/data/sagas.test.js
```diff
@@ -0,0 +1,246 @@
...
+    it('should setup its sub sagas correctly', () => {
+      const gen = saga();
+
+      expect(gen.next()).toMatchSnapshot();
+      expect(gen.next()).toMatchSnapshot();
+      expect(gen.next()).toEqual({ done: true, value: undefined });
+    });
+  });
+
+  describe('handleAddCoupon', () => {
```
davidjoy: But for the action handling sagas... game on!  This test mocks out `apiClient` for the service layer and then runs the actual saga against an action, proving that it dispatches its downstream actions or exceptions properly.  The saga calls the service methods which do their whole thing and get mock data back from the mocked apiClient.  It's a lot of setup, but it really exercises the sagas and services, which is nice.

src/payment/coupon/data/service.js
```diff
@@ -1,23 +1,15 @@
 import pick from 'lodash.pick';
 
-import { handleRequestError } from '../../../common/serviceUtils';
+import { handleRequestError, applyConfiguration } from '../../../common/serviceUtils';
```
davidjoy: "validateConfiguration" has been DRY'd up outside of this.  Using the common one saves me a test. ;)

src/payment/coupon/data/sagas.test.js
```diff
@@ -0,0 +1,246 @@
...
+        addCoupon('DEMO25'),
+      ).toPromise();
+
+      expect(dispatched).toEqual([addCouponBegin(), addCouponFailure('uhoh')]);
+    });
+
+    it('should handle an unexpected error', async () => {
+      const errorResponse = 'oh snap!'; // We don't know how to handle this
+
+      const apiClientPost = jest.fn(() =>
```
abutterworth: Might be nice to put the mock api client code together in a file or at the top of this one as a common object test can use.

src/payment/coupon/data/sagas.test.js
```diff
@@ -0,0 +1,246 @@
...
+
+      const dispatched = [];
+      await runSaga(
+        { dispatch: action => dispatched.push(action) },
+        handleRemoveCoupon,
+        removeCoupon(12345),
+      ).toPromise();
+
+      expect(dispatched).toEqual([
+        removeCouponBegin(),
```
abutterworth: Would it be better to compare the expected objects instead of calling the actions here? e.g. `{ type: REMOVE_COUPON.BEGIN }`

src/payment/coupon/data/sagas.test.js
albemarle: Another option is to make an object of the possible responses up at the top:

```
const responses = {
  success: { ....},
  errorNoKittens: {...},
  errorTooManyKittens: {....}
};```

It does move the responses away from your tests, but depending on the tests, maybe the details don't matter much and just the name is enough to show what you're testing.  (And it serves as a tidy documentation of what you're expecting to see.)

albemarle:
33nxrv

src/payment/coupon/data/sagas.test.js
davidjoy: I'm not sure; doing it that way feels like you're just copy-pasting the guts of the action in here, which feels a bit smelly to me.

src/payment/coupon/data/sagas.test.js
davidjoy: Part of the problem is that if you put the mock setup at the top outside the tests, you need to reset them manually each time in a `beforeEach`.  Seemed more direct to just put the mock creation right there in the test, since it differs for each anyway.  

That said, I'm pulling the response objects out; that feels nice/makes the actual test code look shorter/less daunting.  

I struggled for a while to make this readable and settled on being verbose... thank you both for pushing to make it better. :D

davidjoy: Part of the problem is that if you put the mock setup at the top outside the tests, you need to reset them manually each time in a `beforeEach`.  Seemed more direct to just put the mock creation right there in the test, since it differs for each anyway.  

That said, I'm pulling the response objects out; that feels nice/makes the actual test code look shorter/less daunting.  

I struggled for a while to make this readable and settled on being verbose... thank you both for pushing to make it better. :D

PR #33, opened 2019-06-18
ARCH-853: handle no voucher case

robrap:
ARCH-853


src/payment/data/sagas.js
```diff
@@ -17,7 +17,11 @@ export function* handleFetchBasket() {
   yield put(addCouponBegin());
   const result = yield call(PaymentApiService.getBasket);
   yield put(fetchBasketSuccess(result));
-  yield put(addCouponSuccess(result.voucher.id, result.voucher.code, result.voucher.benefit));
+  if (result.voucher !== undefined) {
```
albemarle: If you have both an if and an else, then often making the condition positive makes it easier to read.

robrap: Done.

PR #35, opened 2019-06-19
refactor: add card to empty basket 

julianajlk:
WIP on UI: Empty Payment Page ([ARCH-932](https://openedx.atlassian.net/browse/ARCH-932))
- Added a card to display the "Your basket is empty" message, without showing the PaymentForm. 

*Side note: 
Used `<FormattedMessage>` as opposed to `formatMessage()` on `renderEmptyMessage()` (L27) because URL links would be nested formatMessage(), which doesn't render `<Hyperlink>` tags. (string - see [explanation](https://github.com/formatjs/react-intl/issues/826)). 

FYI @edx/arch-squad


src/payment/PaymentPage.jsx
```diff
@@ -25,9 +25,12 @@ class PaymentPage extends React.Component {
 
   renderEmptyMessage() {
     return (
-      <p>
-        {this.props.intl.formatMessage(messages['payment.empty.basket'])}
-      </p>
+      <div className="card">
+        <div className="card-body">
+          <h5 className="card-title">{this.props.intl.formatMessage(messages['payment.empty.basket'])}</h5>
+          <p className="card-text">If you attempted to make a purchase, you have not been charged. Return to your <a href="https://courses.edx.org/dashboard">dashboard</a> to try again, or <a href="https://courses.edx.org/support/contact_us">contact edX E-commerce Support</a>.</p>
```
abutterworth: i18n

abutterworth: Checkout this example of getting links inside a FormattedMessage: https://github.com/edx/frontend-app-account/blob/master/src/account-settings/delete-account/BeforeProceedingBanner.jsx#L22

src/payment/PaymentPage.jsx
```diff
@@ -71,11 +74,16 @@ class PaymentPage extends React.Component {
...
+              <div className="col-md-5 pr-md-5 col-basket-summary">
+                {this.renderBasket()}
+              </div> :
+              <div>
+                {this.renderEmptyMessage()}
+              </div>
+            }
             <div className="col-md-7 pl-md-5">
-              <PaymentForm />
+              {isEmpty ? <PaymentForm /> : null}
```
douglashall: Maybe you can have just one `isEmpty` check here instead of multiple.

src/payment/PaymentPage.jsx
douglashall: This text will need to be internationalized as well.

src/payment/PaymentPage.jsx
abutterworth: Instead of rendering the empty message just in the left hand column and then hiding the payment form, let's move this `isEmpty` check so we can render either:
```
 <Row>
    <Col>
      <Basket />
    </Col>
    <Col>
       <Form />
    </Col>
</Row>
```
Or
```
<div>
   <Empty>
</div>
```

src/payment/data/selectors.js
```diff
@@ -11,7 +11,9 @@ export const paymentSelector = (state) => {
   return {
     ...paymentState,
     loaded,
-    isEmpty: loaded && paymentState.products && paymentState.products.length > 0,
+    dashboardURL: state.configuration.LMS_BASE_URL,
+    supportURL: state.configuration.SUPPORT_URL,
+    isEmpty: loaded && paymentState.products && paymentState.products.length > 1,
```
douglashall: This should actually be:
```
isEmpty: loaded && !paymentState.products || paymentState.products.length === 0,
```

douglashall: Sorry, you need some parens in there as well:
```
isEmpty: loaded && (!paymentState.products || paymentState.products.length === 0),
```

src/payment/PaymentPage.jsx
```diff
@@ -24,10 +25,50 @@ class PaymentPage extends React.Component {
...
+                    defaultMessage="dashboard"
+                    description="The message displayed on the redirect to dashboard link"
+                  />
+                </Hyperlink>
+              ),
+              actionLinkTwo: (
+                <Hyperlink destination={supportURL}>
+                  <FormattedMessage
+                    id="payment.empty.basket.supportURL"
+                    defaultMessage="contact edX E-commerce Support"
```
albemarle: I just want to double check this capitalization.

PR #41, opened 2019-06-21
Feedback module for message (errors, success) handling

davidjoy:
To finish this off I need to reproduce the various server error messages and ensure they display correctly, but I think it's code complete.

Tagging @douglashall  as an FYI, as I think it'll be useful for him.  Tagging @abutterworth for review since I'm trying to let Albemarle focus. :)


src/feedback/data/actions.js
```diff
@@ -0,0 +1,20 @@
+export const ADD_MESSAGE = 'ADD_MESSAGE';
+export const REMOVE_MESSAGE = 'REMOVE_MESSAGE';
+
+export const addMessage = (code, message, data, severity, fieldName = null) => ({
```
davidjoy: @douglashall This action is what you're looking for I think.

This should be a comment, probably:

- code and message both optional, but you need one of them.  They're equivalent to error_code and user_message from our server.
- data is any data the client will need to render the message if a code is provided.  i.e., the second argument to `intl.formatMessage`.  
- severity says what color the alert should be.
- fieldName, if provided, identifies the message as belonging to some field (i.e., it came from a field_errors entry)

PR #44, opened 2019-06-26
SDN Check

douglashall:
Check the US government's [list](https://www.treasury.gov/resource-center/sanctions/sdn-list/pages/default.aspx) of Specially Designated Nationals before allowing the purchase flow to continue.

### DIFF
--- a/src/payment/PaymentForm.jsx
+++ b/src/payment/PaymentForm.jsx
@@ -1,7 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import { reduxForm, SubmissionError } from 'redux-form';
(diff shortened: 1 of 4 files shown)


src/payment/PaymentForm.jsx
```diff
@@ -34,6 +39,8 @@ export class PaymentFormComponent extends React.Component {
       throw new SubmissionError(errors);
     }
 
+    sdnCheck(firstName, lastName, city, country);
```
davidjoy: Ideally this call would be made in the sagas to isolate side effects (like making requests) outside the components to keep them more pure.

douglashall:
Thanks @davidjoy! I refactored to move side effects out of the component.

PR #50, opened 2019-07-05
Change frontend names to reflect backend

julianajlk:
Changes to match backend name changes. 
Please note: 
- Coupons are no longer an object and are now an array in the backend, but in the frontend we are only using one coupon `coupons[0]`


src/payment/data/service.js
```diff
@@ -21,27 +22,35 @@ export function configureApiService(newConfig, newApiClient) {
...
-    totalExclDiscount: Number.parseInt(data.total_excl_discount, 10),
-    products: data.products.map(({ image_url: imageURL, title, seat_type: seatType }) => ({
-      imageURL, title, seatType,
+    summaryDiscounts: data.summary_discounts !== null ?
+      Number.parseInt(data.summary_discounts, 10) : null,
+    summaryPrice: Number.parseInt(data.summary_price, 10),
+    products: data.products.map(({
+      image_url: imageUrl,
+      title,
+      certificate_type_display_name: certificateType,
```
nasthagiri: Reminder: This should just be `certificateType`.

src/payment/data/service.js
```diff
@@ -21,27 +22,35 @@ export function configureApiService(newConfig, newApiClient) {
...
+      sku,
+      description,
+      type,
```
nasthagiri: This is now product_type on master.

src/payment/data/service.js
```diff
@@ -21,27 +22,35 @@ export function configureApiService(newConfig, newApiClient) {
...
-      imageURL, title, seatType,
+    summaryDiscounts: data.summary_discounts !== null ?
+      Number.parseInt(data.summary_discounts, 10) : null,
+    summaryPrice: Number.parseInt(data.summary_price, 10),
+    products: data.products.map(({
+      image_url: imageUrl,
+      title,
+      certificate_type_display_name: certificateType,
+      sku,
+      description,
```
nasthagiri: This is now longer being sent.

nasthagiri:
@julianajlk I'm going to go ahead and merge your PR to avoid merge conflicts with PRs from the rest of the team - given that this PR touches many files.

src/payment/coupon/CouponForm.jsx
```diff
@@ -43,7 +41,7 @@ export class CouponForm extends Component {
     switch (errorCode) {
       // Cases that need a `code`
       case 'empty_basket':
-      case 'already_applied_voucher':
+      case 'already_applied_coupon':
```
davidjoy: Has this code been changed on the backend as well?

src/payment/coupon/data/sagas.js
```diff
@@ -23,12 +23,16 @@ export function* handleAddCoupon(action) {
   try {
     const result = yield call(postCoupon, action.payload.code);
     yield put(fetchBasketSuccess(result));
-    if (result.voucher === undefined) {
+    if (result.coupons === undefined) {
```
davidjoy: Do we know that this array will be undefined still?  Or will it be an empty array?

src/payment/data/sagas.js
```diff
@@ -22,10 +22,14 @@ export function* handleFetchBasket() {
   try {
     const result = yield call(PaymentApiService.getBasket);
     yield put(fetchBasketSuccess(result));
-    if (result.voucher === undefined) {
+    if (result.coupons[0] === undefined) {
```
davidjoy: Note that this check differs from above in the coupon sagas.  Up there we check to see if coupons is undefined - here we're assuming it's an array.  Only one of those should be true/they should be consistent... (this is another "How is the backend implemented?" question, but the answer should probably (I hope!) reveal one of these two sagas to have a bug.

julianajlk: Changed to `result.coupons.length === 0`, when coupon is removed, the basket for coupons is an empty array.

src/payment/coupon/data/sagas.js
julianajlk: Empty array, changed to `result.coupons.length === 0`

src/payment/coupon/CouponForm.jsx
julianajlk: Good catch, changed back to `already_applied_voucher`

PR #51, opened 2019-07-05
Show different order details messages per product type.

davidjoy:


src/common/utils.js
```diff
@@ -47,6 +47,38 @@ export function keepKeys(data, whitelist) {
...
+ *   ['first', 'red']
+ * );
+ *
+ * result will be:
+ *
+ * {
+ *   awesome: 'sauce'
+ * }
+ */
+export function getModuleState(state, originalPath) {
```
davidjoy: This is just a helper that I added from one of our other MFEs.

src/common/selectors.js
```diff
@@ -0,0 +1,10 @@
+import { createSelector } from 'reselect';
```
davidjoy: Note this is in the common module - meaning it should ultimately be shared across MFEs.

src/common/utils.test.js
```diff
@@ -115,4 +115,39 @@ describe('keepKeys', () => {
       expect(actionType.RESET).toBe('HOUSE_CATS__START_THE_RACE__RESET');
     });
   });
+
+  describe('getModuleState', () => {
```
davidjoy: Existing test I pulled into the repo.

src/payment/ProductLineItem.jsx
```diff
@@ -49,7 +49,7 @@ class ProductLineItem extends React.PureComponent {
 ProductLineItem.propTypes = {
   imageUrl: PropTypes.string.isRequired,
   title: PropTypes.string.isRequired,
-  certificateType: PropTypes.oneOf(['professional', 'no-id-professional', 'verified', 'honor', 'audit']),
+  certificateType: PropTypes.oneOf(['audit', 'honor', 'verified', 'no-id-professional', 'professional', 'credit']),
```
davidjoy: We were missing 'credit' as a type.

src/payment/__snapshots__/PaymentPage.test.jsx.snap
```diff
@@ -257,9 +257,7 @@ exports[`<PaymentPage /> Renders correctly in various states should render the b
         <h2
           className="section-heading"
         >
-          <span>
-            Order Details
-          </span>
+          Order Details
```
davidjoy: Removed unnecessary span.

src/payment/coupon/data/sagas.test.js
```diff
@@ -50,8 +50,8 @@ describe('saga tests', () => {
             {
               image_url: 'https://prod-discovery.edx-cdn.org/media/course/image/21be6203-b140-422c-9233-a1dc278d7266-941abf27df4d.small.jpg',
               title: 'Introduction to Happiness',
-              certificateType: 'verified',
-              productType: 'Seat',
+              certificate_type: 'verified',
```
davidjoy: This was wrong - must be that the test doesn't actually use it... 🤔

src/payment/data/reducers.js
```diff
@@ -5,6 +5,7 @@ import { reducer as coupon } from '../coupon';
 
 export const basketInitialState = {
   loading: false,
+  products: [],
```
davidjoy: This was necessary because the rest of the code expects it to be an empty array at least, and it's always returned as an empty array from endpoints.  This prevents us from needing downstream isArray checks.

src/payment/data/selectors.js
```diff
@@ -1,27 +1,27 @@
 import { createSelector } from 'reselect';
-import {
-  localeSelector,
-  getCountryList,
-} from '@edx/frontend-i18n';
+import { localeSelector, getCountryList } from '@edx/frontend-i18n';
```
davidjoy: I mildly refactored the contents of this file to be a bit more canonical.  Still needs a little work.

src/payment/order-details/data/constants.js
```diff
@@ -0,0 +1,15 @@
+// Product types
+export const COURSE_ENTITLEMENT_PRODUCT_TYPE = 'Course Entitlement';
```
davidjoy: I'm not thrilled that these constants all live in the order-details module.  I'm thinking about how I think we would ultimately want an 'ecommerce' module that sits beside the payment-page module and contains constants and helpers for ecommerce.  Maybe the API services and sagas live there instead of in payment.  For now... they live here cause this is the only place that really needs them.

src/payment/order-details/data/selectors.js
```diff
@@ -0,0 +1,57 @@
...
+  ENROLLMENT_CODE_PRODUCT_TYPE,
+  COURSE_ENTITLEMENT_PRODUCT_TYPE,
+  ENTITLEMENT_MESSAGE_KEY,
+  ENROLLMENT_CODE_MESSAGE_KEY,
+} from './constants';
+
+export const storePath = ['payment', 'basket'];
+
+const paymentSelector = (state, props) =>
+  getModuleState(state, props.storePath ? props.storePath : storePath);
```
davidjoy: This is a way of making the selectors more modular.  They have a default expectation of where their data is mounted to the redux store, but it can be overridden/configured via the component's props.

src/payment/order-details/OrderDetails.jsx
```diff
@@ -0,0 +1,86 @@
...
+      <p>
+        {this.props.intl.formatMessage(messages[`payment.order.details.course.${this.props.messageType}`])}
+      </p>
+    );
+  }
+
+  renderEnrollmentCodeMessage() {
+    return (
+      <React.Fragment>
+        <p>By purchasing, you and your organization agree to the following terms:</p>
```
albemarle: Do we need to i18n this?

src/payment/order-details/OrderDetails.jsx
```diff
@@ -0,0 +1,86 @@
...
+          <li>
+            Each code will expire in one year from date of purchase or,
+            if earlier, once the course is closed.
+          </li>
+          <li>
+            If a course is not designated as self-paced, you should confirm that a course run is
+            available before expiration.
+          </li>
+          <li>You may not resell codes to third parties.</li>
+          <li>All edX for Business Sales are final and not eligible for refunds.</li>
```
albemarle: Sales -> sales ?

src/payment/order-details/OrderDetails.jsx
davidjoy: Oh junk!  Wow!  I totally forgot.  Of course it does. :open_mouth:

PR #56, opened 2019-07-08
refactor: free checkout

julianajlk:
If the `is_free_basket` boolean is true, it will render just a "Place Order" button that redirects to a free checkout link. When the 100% coupon is removed, the Payment Form is rendered.

### DIFF
--- a/src/payment/PaymentPage.jsx
+++ b/src/payment/PaymentPage.jsx
@@ -18,6 +18,7 @@ import { PageLoading } from '../common';
 import BasketSummary from './BasketSummary';
 import OrderDetails from './order-details';
 import PaymentForm from './PaymentForm';
+import PlaceOrderButton from './PlaceOrderButton';
 import PaymentMethodSelect from './PaymentMethodSelect';
 import ProductLineItems from './ProductLineItems';
 import AlertList from '../feedback/AlertList';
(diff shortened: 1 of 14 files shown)


src/payment/ProductLineItem.jsx
```diff
@@ -35,7 +35,7 @@ class ProductLineItem extends React.PureComponent {
     return (
       <div className="row align-items-center mb-3">
         <div className="col-5">
-          <img className="img-thumbnail" src={imageUrl} alt={title} />
+          <img className="img-thumbnail" src={imageUrl} alt="" />
```
albemarle: Do we need something in the alt attribute for a11y?

src/payment/PaymentPage.jsx
```diff
@@ -90,8 +95,12 @@ class PaymentPage extends React.Component {
           <OrderDetails />
         </div>
         <div className="col-md-7 pl-md-5">
-          <PaymentMethodSelect />
-          <PaymentForm />
+          {isFreeBasket ? <PlaceOrderButton /> : (
```
albemarle: A note for the future: This looks like it changed the way the app works, so that means the semantic commit message should have been fix or feature.  refactor means you didn't change anything visible, you just did the same thing in a different (hopefully better :-) ) way.

PR #61, opened 2019-07-12
Alert message templates

davidjoy:
Described in detail in: https://openedx.atlassian.net/browse/ARCH-1040

NOTE: This PR is against djoy/update_feedback_schema.  Because that branch has been squashed, it won't diff/merge into master very well.

### DIFF
--- a/src/feedback/AlertList.jsx
+++ b/src/feedback/AlertList.jsx
@@ -8,6 +8,20 @@ import { alertListMapStateToProps } from './data/selectors';
 import { removeMessage } from './data/actions';
 
 class AlertList extends Component {
+  renderChild(message) {
+    let child = null;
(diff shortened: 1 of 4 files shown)


src/feedback/AlertList.jsx
```diff
@@ -17,11 +31,15 @@ class AlertList extends Component {
       const {
         id, code, userMessage, messageType, data,
       } = message;
+      if (code != null) {
```
davidjoy: This is a problem - this needs to first check whether we have a normal message for this code.  Only if we don't should we check if we have an alert message template.

davidjoy:
https://openedx.atlassian.net/browse/ARCH-1040 was completed. Closing!

PR #62, opened 2019-07-12
feat: add apple pay

abutterworth:
TODO:

- [x] Add apple config to global config
- [x] i18n
- [ ] Verify working (will do on stage)
- [x] Redirect to receipt on success
- [x] SDN check if needed
- [x] Pass good error messages through to UI message
- [ ] remove console logs (will remove in a subsequent PR)
- [x] tests

Screen Shot 2019-07-12 at 4 57 06 PM


douglashall:
How do I get the ApplePay button to show up in the browser?

src/payment/ApplePay.jsx
```diff
@@ -0,0 +1,54 @@
...
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { injectIntl, intlShape } from '@edx/frontend-i18n';
+
+import { basketSelector } from './data/selectors';
+import { addMessage, MESSAGE_TYPES } from '../feedback';
+import { ApplePayButton, redirectToReceipt } from './apple-pay';
+import messages from './ApplePay.messages';
+
+function ApplePay(props) {
```
douglashall: Should this component be located under the `apple-pay` directory?

abutterworth: It could be – or could be folded directly into ApplePayButton. This is basically the component - container split. I'll move in the directory and rename it ApplePayButtonContainer or something. wdyt?

douglashall: Yeah, I think I like that. I'm only bringing this up because I think I'll go with a similar layout for the PayPal button which I'm implementing now.

src/payment/apple-pay/ApplePayButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+    if (!this.state.canMakePayments || this.props.totalAmount === undefined) return null;
+
+    const other = { ...this.props };
+
+    delete other.totalAmount;
+    delete other.onPaymentBegin;
+    delete other.onPaymentComplete;
+    delete other.onMerchantValidationFailure;
+    delete other.onPaymentAuthorizationFailure;
+    delete other.onPaymentCancel;
```
douglashall: What's going on here with these `delete`s?

abutterworth: I want users to pass any html attribute through if they want to, but these props are not valid. I could use lodash's .omit() instead but opted not to use a library

PR #64, opened 2019-07-15
refactor: add scroll to error on submit

julianajlk:
Added `scrollToErrors` where onSubmit of the Payment form, it scrolls to the error message. If there are multiple errors in the form's input fields, scrolls to first error. 
[ARCH-939](https://openedx.atlassian.net/browse/ARCH-939).

### DIFF
--- a/src/payment/PaymentForm.test.jsx
+++ b/src/payment/PaymentForm.test.jsx
@@ -18,16 +18,20 @@ const storeMocks = {
 configureI18n(configuration, messages);
 
 describe('<PaymentForm />', () => {
+  let wrapper;
+  let paymentForm;
+  beforeEach(() => {
+    wrapper = mount((
+      <IntlProvider locale="en">
+        <Provider store={mockStore(storeMocks.defaultState)}>
+          <PaymentForm handleSubmit={() => {}} />
+        </Provider>
+      </IntlProvider>
+    ));
+    paymentForm = wrapper.find(PaymentFormComponent).first().instance();
+  });
   describe('getRequiredFields', () => {
     it('returns expected required fields', () => {
-      const wrapper = mount((
-        <IntlProvider locale="en">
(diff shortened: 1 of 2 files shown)


src/payment/PaymentForm.test.jsx
```diff
@@ -174,14 +176,29 @@ describe('<PaymentForm />', () => {
...
-        invalid: 'This field is required',
+        lastName: 'This field is required',
       };
-
       expect(paymentForm.validateRequiredFields(values)).toEqual(expectedErrors);
     });
   });
+  describe('scrollToError', () => {
+    it('scrolls to the input name of the first error', () => {
+      const wrapper = mount((
```
robrap: Consider moving the setup of `paymentForm` to a setup function like beforeEach(), or some other shared function.  See https://jestjs.io/docs/en/setup-teardown.html

julianajlk: Great, added `let wrapper;
    beforeEach(() => {
      wrapper = mount((
        <IntlProvider locale="en">
          <Provider store={mockStore(storeMocks.defaultState)}>
            <PaymentForm handleSubmit={() => {}} />
          </Provider>
        </IntlProvider>
      ));
    });`

src/payment/PaymentForm.test.jsx
```diff
@@ -18,16 +18,20 @@ const storeMocks = {
 configureI18n(configuration, messages);
 
 describe('<PaymentForm />', () => {
+  let wrapper;
+  let paymentForm;
+  beforeEach(() => {
+    wrapper = mount((
```
robrap: Note: I think this could have been `let wrapper = ...` because I think it was only used here, but I could be wrong.

julianajlk: Changed to `const wrapper`

PR #67, opened 2019-07-17
add first purchase discount to mfe

MatthewPiatetsky:
I still need to write tests for this, but I wanted to validate my approach at a high level in the meantime

### DIFF
--- a/src/payment/data/sagas.js
+++ b/src/payment/data/sagas.js
@@ -53,8 +53,54 @@ export function* performBasketOperation(operation, ...operationArgs) {
   }
 }
 
+/**
+ * Many of the handlers here have identical error handling, and are also all processing the same
+ * sort of response object (a basket).  This helper is just here to dedupe that code, since its
+ * all identical.
+ */
+export function* catchBasketError(handler) {
+  if (yield isBasketProcessing()) return;
+
+  try {
+    yield call(handler);
+  } catch (error) {
+    yield call(handleErrors, error, true);
+    if (error.basket) {
+      yield put(basketDataReceived(error.basket));
+    }
+  } finally {
+    yield put(basketProcessing(false));
+  }
+}
+
+export function* handleDiscountCheck() {
+  const basket = yield select(state => state.payment.basket);
+  if (basket.products.length === 1) {
+    const [product] = basket.products;
+    const { courseKey } = product;
+    const discount = yield call(
+      PaymentApiService.getDiscountData,
+      courseKey,
+    );
+    if (discount.discount_applicable) {
+      const result = yield call(
+        PaymentApiService.getBasket,
+        discount.jwt,
+      );
+      yield put(basketDataReceived(result));
+      yield call(handleMessages, result.messages, true);
+    }
+  }
+}
+
 export function* handleFetchBasket() {
-  yield call(performBasketOperation, PaymentApiService.getBasket);
+  yield call(catchBasketError, function* processFetchBasket() {
+    yield put(basketProcessing(true));
+    const result = yield call(PaymentApiService.getBasket);
+    yield put(basketDataReceived(result));
+    yield call(handleMessages, result.messages, true);
+    yield call(handleDiscountCheck);
+  });
   yield put(fetchBasket.fulfill());
 }
 
--- a/src/payment/data/sagas.js
+++ b/src/payment/data/sagas.js
@@ -88,7 +88,7 @@ export function* handleDiscountCheck() {
         discount.jwt,
       );
       yield put(basketDataReceived(result));
-      yield call(handleMessages, result.messages, true);
+      yield call(handleMessages, result.messages, false);
     }
   }
 }
--- a/src/payment/data/service.js
+++ b/src/payment/data/service.js
@@ -86,9 +86,10 @@ function handleBasketApiError(requestError) {
   }
 }
 
-export async function getBasket() {
+export async function getBasket(discountJwt) {
+  const discountJwtArg = typeof discountJwt !== 'undefined' ? `?discount_jwt=${discountJwt}` : '';
   const { data } = await apiClient
-    .get(`${config.ECOMMERCE_BASE_URL}/bff/payment/v0/payment/`)
+    .get(`${config.ECOMMERCE_BASE_URL}/bff/payment/v0/payment/${discountJwtArg}`)
     .catch(handleBasketApiError);
 
   return transformResults(data);
@@ -119,4 +120,14 @@ export async function deleteCoupon(id) {
     .delete(`${config.ECOMMERCE_BASE_URL}/bff/payment/v0/vouchers/${id}`)
     .catch(handleBasketApiError);
   return transformResults(data);
+}
+
+export async function getDiscountData(courseKey) {
+  const { data } = await apiClient.get(
+    `${config.LMS_BASE_URL}/api/discounts/course/${courseKey}`,
+    {
+      xhrFields: { withCredentials: true },
+    },
+  );
+  return data;
 }
--- a/README.rst
+++ b/README.rst
@@ -154,6 +154,7 @@ described below:
   ``127.0.0.1 local.stage.edx.org``.
 
 - Log into stage: `https://courses.stage.edx.org/login <https://courses.stage.edx.org/login>`_.
+- Run `npm install` in this project directory
 - Start the frontend's dev server in staging mode:
 
   ``npm run start:stage``
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
     "@babel/cli": {
       "version": "7.5.5",
       "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.5.5.tgz",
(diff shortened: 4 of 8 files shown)


src/payment/data/service.js
```diff
@@ -70,7 +70,6 @@ export function transformResults(data) {
 
   if (results.offers != null) {
     results.offers = results.offers
-      .filter(({ provider }) => provider !== null)
```
MatthewPiatetsky: enterprise offers have a provider, dynamic conditional offers (added for first purchase discount) do not

src/payment/data/service.js
```diff
@@ -80,9 +79,10 @@ export function transformResults(data) {
   return results;
 }
 
-export async function getBasket() {
+export async function getBasket(discountJwt) {
+  const discountJwtArg = typeof discountJwt !== 'undefined' ? `?discount_jwt=${discountJwt}` : '';
   const { data } = await apiClient
-    .get(`${config.ECOMMERCE_BASE_URL}/bff/payment/v0/payment/`)
+    .get(`${config.ECOMMERCE_BASE_URL}/bff/payment/v0/payment/${discountJwtArg}`)
```
MatthewPiatetsky: the first purchase discount offer gets applied when a discount jwt is sent as an argument

src/payment/data/service.js
```diff
@@ -114,3 +114,13 @@ export async function deleteCoupon(id) {
     .catch(handleBasketApiError);
   return transformResults(data);
 }
+
+export async function checkBasketForDiscount(courseKey) {
```
MatthewPiatetsky: in ecommerce (and mfe) basket loads, async request goes to lms to check if a discount should be  applied, if it is true, basket is recreated server side and page reloads (or re-renders in the case of the mfe)

src/payment/order-summary/Offers.jsx
```diff
@@ -16,23 +16,30 @@ Benefit.propTypes = {
   benefitValue: PropTypes.number.isRequired,
 };
 
+function Message({ benefitType, benefitValue, provider }) {
```
MatthewPiatetsky: enterprise offers and the first purchase offer have different messages

src/payment/data/sagas.js
```diff
@@ -19,10 +19,38 @@ import { handleErrors, handleMessages } from '../../feedback';
...
+        courseKey,
+      );
+      if (discount.discount_applicable) {
+        try {
+          yield put(fetchBasket.request());
+          result = yield call(
+            PaymentApiService.getBasket,
+            discount.jwt,
+          );
+          yield put(fetchBasket.success(result));
```
MatthewPiatetsky: when we need to update the page after the discount is applied @abutterworth just suggested one option of just copying the code to be run a second time
but if we need to pull something out to a separate function to be more DRY that's fine too

src/payment/data/sagas.js
```diff
@@ -19,10 +19,38 @@ import { handleErrors, handleMessages } from '../../feedback';
...
     yield put(fetchBasket.request());
-    const result = yield call(PaymentApiService.getBasket);
+    let result = yield call(PaymentApiService.getBasket);
     yield put(fetchBasket.success(result));
     yield put(basketDataReceived(result));
     yield call(handleMessages, result.messages, true);
+    if (result.products.length === 1) {
+      const [product] = result.products;
+      const { courseKey } = product;
+      const discount = yield call(
```
davidjoy: I'm wondering if we can off-load this work to the API itself so that the first time we call getBasket we just get all the information we need.  i.e., the original call to getBasket on line 22 would just include whether or not a first purchase discount is available.

Failing that, it might be cleaner to create a service method that encapsulates this logic/conditional calling of checkBasketForDiscount so that we can keep it out of the saga.  That would mean that on line 22 we'd come back here with all the data we need, and wouldn't have to worry about DRYing up all the logic here/re-dispatching all the redux actions and all that.

src/payment/data/service.js
davidjoy: Nit: we've been trying to prefix these service method names with an HTTP verb/have their name represent the endpoint, for what it's worth.  Something like `getCourseDiscount`, maybe?

src/payment/data/sagas.js
MatthewPiatetsky: for point 1, i believe @robrap has more context on this, but i believe the decision was not to have the data available on the first request, so that we wouldn't need to make a synchronous request to the lms from ecommerce

MatthewPiatetsky: for point 2, i think your suggestion is to get all the changes we need on line 22 before we complete rendering the basket the first time.  
i believe this would then essentially be a synchronous request, which i think @robrap said we can't do? again, i don't have the full context on this, but i think @robrap does
(and @emma-green)

src/payment/order-summary/Offers.jsx
```diff
@@ -16,23 +16,30 @@ Benefit.propTypes = {
...
+  if (provider) {
+    return [<Benefit benefitType={benefitType} benefitValue={benefitValue} />, ' discount provided by', provider];
```
davidjoy: If I'm reading this code right, 'discount provided by' and 'discount for your first upgrade applied' aren't localized anymore...  Might be more clearly factored by taking the conditional in `Message` and just inserting it into `Offer` directly and choosing one of two `<FormattedMessage>` components to insert into the paragraph tag.

Something like (I have not tested this code):

```
function Offer({
  benefitType, benefitValue, provider,	  benefitType, benefitValue, provider,
}) {
  let message = null;
  if (provider) {
    message = <FormattedMessage
      id="payment.summary.discount.offer"
      defaultMessage="{benefit} discount provided by {provider}."
      description="A description of a discount offer applied to a basket."
      values={{
          benefit: (
            <Benefit
              benefitType={benefitType}
              benefitValue={benefitValue}
            />
          ),
          provider,	
        }}
      />;
    } else {
      message = <FormattedMessage
        id="payment.summary.discount.offer"
        defaultMessage="{benefit} discount for your first upgrade applied."
        description="A description of a discount offer applied to a basket for your first upgrade."
        values={{
          benefit: (
            <Benefit
              benefitType={benefitType}
              benefitValue={benefitValue}
            />
          ),
        }}
      />;
    }
  return (
    <p className="m-0 text-muted" key={`${benefitValue}-${provider}`}>	    <p className="m-0 text-muted" key={`${benefitValue}-${provider}`}>
      {message}
    </p>
  );
}
```

davidjoy: (obviously the <FormattedMessage> itself could be DRY'd up a bit, but this way we can just do it using the existing components)

MatthewPiatetsky: I'm not familiar with how formatted message works, this is the method that does the internationalization here?
Is the issue that because the text being inserted is a variable/react component, it won't be internationalized?  
If so, could I just call FormattedMessage directly on the strings that are currently not internationalized in their current location?

davidjoy: Usages of `<FormattedMessage />` are grepped for by a script at build time, and the defaultMessage, description, and ID are bundled off to Transifex for translation along with files of the pattern `*.messages.js`.  So any text that's not part of one - i.e, the text above in `Message`, won't be internationalizable.

The rest of my suggestion is just a more canonical/standard way of factoring the code in light of the above need to make things internationalizable, IMHO.

src/payment/data/sagas.js
davidjoy: Apologies, this is hard to explain in text. :)

For point 2, I'm suggesting that the logic of 1) calling `getBasket`, 2) then `checkBasketForDiscount`, 3) then possibly `getBasket` again, could be extracted from this saga and hidden behind a single service method call on line 22, something like:

```
let result = yield call(PaymentApiService.getBasketWithPossibleDiscount) // terrible name`
```

And then the saga here would effectively stay the same as it originally was, without any knowledge that those extra calls were being made. Then in your new service method you could write the logic without all the extra syntax of this saga:

```
async function getBasketWithPossibleDiscount() {
  let result = await getBasket();
  if (result.products.length === 1) {
    const [product] = result.products;
    const { courseKey } = product;
    const discount = await checkBasketForDiscount(courseKey);
    if (discount.discount_applicable) {
      result = await getBasket(discount.jwt);
    }
    return result;
}
```

Let me know if that doesn't make sense - happy to try to explain with spoken words!

MatthewPiatetsky: thanks! i tried to implement what you had in slack.
omitting this after the second getBasket call seems to work, but what does this statement do?
yield put(fetchBasket.success(result));

src/payment/order-summary/Offers.jsx
MatthewPiatetsky: makes sense
i've tried to implement your suggestion above
it seems like this would be more total lines than if the message was factored out into two formatted messages in its own function, but if this is the more canonical way to do it then that's fine

src/payment/data/service.js
MatthewPiatetsky: updated to getDiscountData

PR #69, opened 2019-07-18
refactor: add event parity to events

julianajlk:
WIP on event parity `onClick` and `onSubmit` of "place order" buttons both for PaymentForm and Free Basket, [ARCH-941](https://openedx.atlassian.net/browse/ARCH-941). 

- Blocked on: how to test `sendTrackEvent()` manually?


src/payment/PaymentForm.jsx
```diff
@@ -104,6 +105,21 @@ export class PaymentFormComponent extends React.Component {
     return requiredFields;
   }
 
+  handleTrackEvent() {
```
abutterworth: I would change this name to something like `handleSubmitButtonClick` or something. Probably better to describe the event it handles rather than what it will do when that even happens.

src/payment/PaymentForm.test.jsx
```diff
@@ -5,6 +5,7 @@ import configureMockStore from 'redux-mock-store';
 import { mount } from 'enzyme';
 import { SubmissionError } from 'redux-form';
 import { IntlProvider, configure as configureI18n } from '@edx/frontend-i18n';
+import * as analytics from '@edx/frontend-analytics';
```
abutterworth: If you aren't actually using analytics and just want to mock it. I'd suggest not importing the package and instead stub it out.
`const analytics = { sendTrackEvent: jest.fn() }`

src/payment/apple-pay/ApplePayButton.test.jsx
```diff
@@ -2,6 +2,7 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import renderer from 'react-test-renderer';
+import * as analytics from '@edx/frontend-analytics';
```
abutterworth: Same as above.

src/payment/PaymentForm.test.jsx
julianajlk: Used `jest.mock('@edx/frontend-analytics', () => ({
  sendTrackEvent: jest.fn(),
}));`

PR #71, opened 2019-07-19
feat: add optimizely snippet to top of head

abutterworth:
The webpack html template will now produce like below if `process.env.OPTIMIZELY_PROJECT_ID` exists. (Companion PR: https://github.com/edx/edx-internal/pull/1010)

```
<!DOCTYPE html>
<html lang="en-us">
<head>
    <script src="https://cdn.optimizely.com/js/1234.js"></script>
    ...
```

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,9 +1,12 @@
 <!doctype html>
 <html lang="en-us">
   <head>
-      <title>Payment | edX</title>
-      <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


marcotuts:
Is this something we should be adding via Segment.io instead of directly?

abutterworth:
@marcotuts We need optimizely on the page to enable manipulating the page for a/b testing. (on top of tracking).

PR #74, opened 2019-07-22
feat: send errors to NR and add fallback error message

abutterworth:
Screen Shot 2019-07-22 at 9 22 36 AM

Log info and errors to New Relic on request errors. Update `handleErrors` to provide a fallback error message rather than bailing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1561,9 +1561,9 @@
       }
     },
     "@edx/frontend-logging": {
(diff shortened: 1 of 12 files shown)


marcotuts:
thoughts on language around refreshing / trying again? this is fine to merge as is, but I'm wondering what options we would have to update this in case it applies generally for other 404 type cases.

PR #77, opened 2019-07-22
Fail commit on npm audit vulnerabilities

davidjoy:
`npm audit` kindly returns a non-zero exit status on any failure.

Note that it also has an `--audit-level` parameter we can choose to use if we want more granular control over what constitutes a failure.  For now it'll fail on any vulnerability.

https://docs.npmjs.com/cli/audit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ script:
   - make validate-no-uncommitted-package-lock-changes
   - npm run i18n_extract
   - npm run lint
+  - npm audit
   - npm run test
   - npm run build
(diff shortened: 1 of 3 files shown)


davidjoy:
As written, no, it only runs on precommit.  We can fix that if we'd like.

robrap:
I think I would make it fail for now, and if and when it becomes a problem, we can adjust with `--audit-level` as needed.

davidjoy:
I added the audit to travis earlier.  Think this PR should be ready to go.

PR #99, opened 2019-07-23
Tests backfill - first pass

davidjoy:
Tests for generateAndSubmitform and the PlaceOrderButton, bringing both to 100% coverage

### DIFF
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -113,21 +113,19 @@ export class AsyncActionType {
 }
 
 export function generateAndSubmitForm(url, params = {}) {
-  const form = document.createElement('form');
+  const form = global.document.createElement('form');
   form.method = 'POST';
   form.action = url;
 
   Object.keys(params).forEach((key) => {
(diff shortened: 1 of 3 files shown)


src/common/utils.js
```diff
@@ -113,21 +113,19 @@ export class AsyncActionType {
 }
 
 export function generateAndSubmitForm(url, params = {}) {
-  const form = document.createElement('form');
+  const form = global.document.createElement('form');
   form.method = 'POST';
   form.action = url;
 
   Object.keys(params).forEach((key) => {
-    if (Object.prototype.hasOwnProperty.call(params, key)) {
```
albemarle: We no longer care if it's an own property?

davidjoy: We do, but `Object.keys` filters by "own, enumerable properties" already, so it's a duplicative check.

davidjoy: Anticipating this question I made a note of _why_ I made this change... in the commit message that doesn't really get shown in the PR in an obvious place. :)

PR #110, opened 2019-07-23
feat: add bulk enrollment

abutterworth:
Screen Shot 2019-07-24 at 9 55 14 AM


src/payment/OrderSummary.jsx
```diff
@@ -0,0 +1,242 @@
...
+function SummaryTable({ price }) {
+  if (price === undefined) return null;
+
+  return (
+    <div className="summary-row d-flex">
+      <span className="flex-grow-1">
+        <FormattedMessage
+          id="payment.summary.table.label.price"
+          defaultMessage="Price"
+          description="Label for price excluding discount line on order summary table"
```
albemarle: I don't think we need to specify "excluding discount line"-- I think that makes it hard for a non-native English speaker  to understand, and probably all they need to know is that it's the price.

PR #116, opened 2019-07-24
fix: add coupon sendTrackEvent

julianajlk:
Added ecommerce event for coupons `edx.bi.ecommerce.basket.voucher_applied`.
[ARCH-941](https://openedx.atlassian.net/browse/ARCH-941).


src/payment/coupon/CouponForm.test.jsx
```diff
@@ -91,6 +96,27 @@ describe('CouponForm', () => {
     expect(addCoupon).toHaveBeenCalledWith('DEMO25');
   });
 
+  it('sends coupon track data on form submit', () => {
```
albemarle: Thank you for doing tests as you go along.  :)

PR #119, opened 2019-07-24
feat: update basket data on all requests

abutterworth:
Also pass through get parameter to first getBasket request.

This will need a refactor in the future.

### DIFF
--- a/src/payment/coupon/data/sagas.js
+++ b/src/payment/coupon/data/sagas.js
@@ -2,14 +2,13 @@ import { call, put, takeEvery } from 'redux-saga/effects';
 import {
   ADD_COUPON,
   addCouponBegin,
-  addCouponSuccess,
   removeCouponBegin,
   removeCouponSuccess,
   REMOVE_COUPON,
   addCouponFailure,
   removeCouponFailure,
 } from './actions';
-import { fetchBasketSuccess } from '../../data/actions';
+import { handleBasketResult } from '../../data/sagas';
 import { deleteCoupon, postCoupon } from './service';
 
 import { handleErrors, handleMessages } from '../../../feedback';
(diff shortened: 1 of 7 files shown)


src/payment/coupon/data/sagas.js
```diff
@@ -39,10 +32,13 @@ export function* handleRemoveCoupon(action) {
   yield put(removeCouponBegin());
   try {
     const result = yield call(deleteCoupon, action.payload.id);
-    yield put(fetchBasketSuccess(result));
+    yield call(handleBasketResult, result);
     yield put(removeCouponSuccess(result));
```
nasthagiri: Does this remove a coupon success message?  Should that happen before handling the newly returned result?

src/payment/coupon/data/sagas.js
```diff
@@ -18,18 +17,12 @@ export function* handleAddCoupon(action) {
   yield put(addCouponBegin());
   try {
     const result = yield call(postCoupon, action.payload.code);
-    yield put(fetchBasketSuccess(result));
+    yield call(handleBasketResult, result);
     yield call(handleMessages, result.messages, true);
```
nasthagiri: Should `handleMessages` just be part of `handleBasketResult`?

src/payment/coupon/data/service.js
```diff
@@ -1,6 +1,7 @@
 import pick from 'lodash.pick';
 
-import { handleRequestError, applyConfiguration } from '../../../common/serviceUtils';
+import { applyConfiguration } from '../../../common/serviceUtils';
```
nasthagiri: Is this import of `applyConfiguration` used?

src/payment/coupon/data/sagas.js
nasthagiri: Should removal of old messages just be part of `handleBasketResult`?

src/payment/data/service.js
```diff
@@ -36,6 +39,26 @@ export function configureApiService(newConfig, newApiClient) {
     }
     return response;
   });
+
+  const decodeURLParams = (search) => {
```
nasthagiri: Future thought:
I don't think it's necessary right now, but in future, we may want to forward other types of params as well.  For now, it seems UTM params aren't needed to be forwarded.

src/payment/data/sagas.js
```diff
@@ -20,24 +20,32 @@ import { saga as cybersourceSaga } from '../cybersource';
 import { saga as payPalSaga } from '../paypal';
 import { handleErrors, handleMessages } from '../../feedback';
 
+export function* handleBasketResult(basketResult) {
```
nasthagiri: Consider going through one more pass on DRY-ing further by including message displaying and error handling in here as well?

src/payment/coupon/data/sagas.js
abutterworth: We handle messages on success and errors on failure, but both need to populate the basket.

src/payment/coupon/data/service.js
abutterworth: Yes it's in there.

src/payment/data/sagas.js
abutterworth: There's some complication around how errors and messages are handled that make this less than straightforward. I'll address in a refactor later.

PR #120, opened 2019-07-25
fix: a11y

julianajlk:
Updated headings to include `aria-level=""` instead of changing the heading size on CSS.


src/payment/BasketSummary.jsx
```diff
@@ -0,0 +1,185 @@
...
+
+TotalTable.propTypes = {
+  orderTotal: PropTypes.number,
+};
+TotalTable.defaultProps = {
+  orderTotal: undefined,
+};
+
+
+function BasketSummary(props) {
```
abutterworth: This whole file has been renamed to OrderSummary

PR #122, opened 2019-07-25
feat: add a loading skeleton for cart section

abutterworth:
Screen Shot 2019-07-25 at 3 12 14 PM

### DIFF
--- a/src/payment/PaymentForm.jsx
+++ b/src/payment/PaymentForm.jsx
@@ -24,6 +24,8 @@ export class PaymentFormComponent extends React.Component {
   }
 
   onSubmit = (values) => {
+    if (this.props.loading) return;
+
     const requiredFields = this.getRequiredFields(values);
     const {
       firstName,
@@ -184,6 +186,7 @@ export class PaymentFormComponent extends React.Component {
     const {
       handleSubmit,
       submitting,
+      loading,
       orderType,
     } = this.props;
 
@@ -200,13 +203,19 @@ export class PaymentFormComponent extends React.Component {
         <CardDetails submitting={submitting} />
         <div className="row justify-content-end">
           <div className="col-lg-6 form-group">
-            <button type="submit" className="btn btn-primary btn-lg btn-block" disabled={submitting} onClick={this.handleSubmitButtonClick}>
-              <FormattedMessage
-                id="payment.form.submit.button.text"
-                defaultMessage="Place Order"
-                description="The label for the payment form submit button"
-              />
-            </button>
+            {
+              loading ? (
+                <div className="skeleton btn btn-block btn-lg rounded-pill">&nbsp;</div>
+              ) : (
+                <button type="submit" className="btn btn-primary btn-lg btn-block" disabled={submitting || loading} onClick={this.handleSubmitButtonClick}>
+                  <FormattedMessage
+                    id="payment.form.submit.button.text"
+                    defaultMessage="Place Order"
+                    description="The label for the payment form submit button"
+                  />
+                </button>
+              )
+            }
           </div>
         </div>
       </form>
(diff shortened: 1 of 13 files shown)


src/payment/PaymentForm.jsx
```diff
@@ -200,13 +203,19 @@ export class PaymentFormComponent extends React.Component {
...
           <div className="col-lg-6 form-group">
-            <button type="submit" className="btn btn-primary btn-lg btn-block" disabled={submitting} onClick={this.handleSubmitButtonClick}>
-              <FormattedMessage
-                id="payment.form.submit.button.text"
-                defaultMessage="Place Order"
-                description="The label for the payment form submit button"
-              />
-            </button>
+            {
+              loading ? (
```
robrap: Nit: this doesn't have to change, but I'm used to using `A ? B : C` for short one-liners.  For multi-line conditionals, I prefer `if/else`, but maybe I'm just old, or not up to frontend best practices?

src/payment/PaymentMethodSelect.jsx
```diff
@@ -24,15 +26,23 @@ function PaymentMethodSelect({ intl }) {
             alt={intl.formatMessage(messages['payment.page.method.type.credit'])}
           />
         </button>
-        <PayPalButton />
-        <ApplePayButtonContainer />
+
+        <PayPalButton
+          className={classNames({ 'skeleton-pulse': loading })}
```
robrap: What does `skeleton-pulse` mean?

src/payment/apple-pay/ApplePayButton.jsx
```diff
@@ -14,6 +15,8 @@ export default class ApplePayButton extends React.Component {
   }
 
   handleClick = () => {
+    if (this.props.disabled || this.props.totalAmount === undefined) return;
```
robrap: Can this be different from `canMakePayments` below?

abutterworth: canMakePayments is the browser capability. May be worth combining these into a single boolean.

src/payment/PaymentMethodSelect.jsx
abutterworth: It's a class name for making an element fade in and out – you can see the animation in the screenshot (obviously :P)

abutterworth:
> 1. You had mentioned that on mobile we might need to scroll to the right place when things appear?

Yes, I think we should merge this and come back to it. It may be less of an issue if we're filling in space with these "skeletons".

> 2. As mentioned in slack, what are the i18n concerns with the skeletons used for loading?, and

Unsure, will check.

> 3. Will there be any visual indicator of a spinner or anything to make it clear that the skeleton is because something is still happening?  Not sure when and if you want to bring in UX on this.

The skeletons fade in and out – you can see an almost identical implementation in JIRA. I prefer these over any spinner. IMO they're just as effective and less distracting.

src/payment/apple-pay/ApplePayButton.jsx
robrap: Seems like a helper that combines all of this would help avoid errors, but I'm not 100% sure.  I'd recommend doing something, even if in a follow-up PR.  Thanks.

abutterworth: 👍

src/payment/PaymentForm.jsx
abutterworth: In JSX I think this the only way to do it. I could pull it all out to a function instead, which is probably the preferred approach when things start to get big.

robrap: I see.  Thanks.  Yes - it could be a function.

PR #123, opened 2019-07-25
feat: write out APPLE_DEVELOPER_MERCHANT_ID_DOMAIN_ASSOCIATION to fil…

abutterworth:
feat: write out APPLE_DEVELOPER_MERCHANT_ID_DOMAIN_ASSOCIATION to file for verification.

Writes a file to `dist/.well-known/apple-developer-merchantid-domain-association.txt` as described here: https://developer.apple.com/documentation/apple_pay_on_the_web/maintaining_your_environment . Companion PR: https://github.com/edx/edx-internal/pull/1024


build.js
```diff
@@ -0,0 +1,35 @@
+const webpack = require('webpack'); // eslint-disable-line import/no-extraneous-dependencies
+const config = require('./webpack/webpack.prod.config.js');
```
robrap: Is `build.js` still a prod specific file because it imports `webpack.prod.config.js`, but isn't named like it is a prod specific file?  Or, is it conditionally prod-like?
Can you either add a comment here, or update the name of this file, as appropriate depending on the answer?

abutterworth: Good call.

PR #132, opened 2019-07-26
fix(prices): fix display of absolute value coupons

albemarle:
The style of FormattedMessage that takes a child function can only handle text arguments to that function, not elements.

### DIFF
--- a/src/payment/coupon/CouponForm.jsx
+++ b/src/payment/coupon/CouponForm.jsx
@@ -96,19 +96,19 @@ export class CouponForm extends Component {
 
     if (benefitType === 'Absolute') {
       return (
-        <FormattedMessage
-          id="payment.coupon.benefit.absolute"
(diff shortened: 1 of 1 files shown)


src/payment/coupon/CouponForm.jsx
```diff
@@ -96,19 +96,19 @@ export class CouponForm extends Component {
...
-          defaultMessage="Coupon {code} applied for {amount} off"
-          description="A coupon has been applied for a fixed currency discount, like $10.  Currency symbol will already be provided."
-          values={{
-            code,
-            amount: (
-              <LocalizedPrice amount={benefitValue} />
-            ),
-          }}
-        >
-          {renderMuted}
```
robrap: I would either:
1. Delete `renderMuted` and switch the other usage to match this pattern, or
2. Either inline or move `renderMuted` to its lone usage, rather than the top of the file, since it is no longer shared.

PR #134, opened 2019-07-26
refactor: coupon module away

abutterworth:
- Refactor coupon module to no longer maintain a copy of the coupon state.
- Leverage redux-saga-routines library to eliminate lots of boilerplate


src/payment/__snapshots__/PaymentPage.test.jsx.snap
```diff
@@ -430,16 +412,17 @@ exports[`<PaymentPage /> Renders correctly in various states should render all c
                 className="h6 d-block"
                 htmlFor="couponField"
               >
-                Add coupon code (optional)
+                <span>
```
albemarle: How much do we care about these extra span tags?  If you use the version of FormattedMessage that takes a function as its child, you should be able to coax it not to use any tags at all, I think.

src/payment/__snapshots__/PaymentPage.test.jsx.snap
```diff
@@ -6438,37 +6423,19 @@ exports[`<PaymentPage /> Renders correctly in various states should render the b
...
-              disabled={false}
+              className="btn btn-link display-inline p-0 pl-3 border-0"
               onBlur={[Function]}
               onClick={[Function]}
               onKeyDown={[Function]}
               type="submit"
             >
-              Apply
+              <span>
+                Remove
```
albemarle: Hmm, why did Apply change to Remove?

src/payment/__snapshots__/PaymentPage.test.jsx.snap
```diff
@@ -129,37 +129,19 @@ exports[`<PaymentPage /> Renders correctly in various states should render a fre
...
```
albemarle: Apply became Remove?

src/payment/data/sagas.js
```diff
@@ -2,77 +2,99 @@ import { all, call, put, takeEvery } from 'redux-saga/effects';
...
+    }
+  } finally {
+    yield put(addCoupon.fulfill());
+  }
+}
+
+export function* handleRemoveCoupon({ payload }) {
+  try {
+    yield put(removeCoupon.request());
+    const result = yield call(PaymentApiService.deleteCoupon, payload.code);
```
nasthagiri: Nit: rename `deleteCoupon` to `removeCoupon` to keep consistent naming.

src/payment/data/sagas.js
```diff
@@ -2,77 +2,99 @@ import { all, call, put, takeEvery } from 'redux-saga/effects';
...
+    if (error.basket) {
+      yield put(basketDataReceived(error.basket));
     }
-    yield call(handleErrors, e, true);
+  } finally {
+    yield put(removeCoupon.fulfill());
   }
 }
 
 export function* handleUpdateEnrollmentCodeQuantity({ payload }) {
```
nasthagiri: NIt: rename this to `handleUpdateQuantity` to have consistent naming - since this PR also renames other code to just `Quantity` instead of `EnrollmentCodeQuantity`.

src/payment/__snapshots__/PaymentPage.test.jsx.snap
```diff
@@ -6438,37 +6444,19 @@ exports[`<PaymentPage /> Renders correctly in various states should render the b
...
             onSubmit={[Function]}
           >
-            <div
-              className="form-group mb-0 mr-2"
-            >
-              <label
-                className="h6 d-block"
-                htmlFor="couponField"
-              >
-                Add coupon code (optional)
```
nasthagiri: Did the default state change from Adding a Coupon mode to a Removal of Coupon mode?  You may want to double-check why the snapshot test changed in this way.

src/payment/__snapshots__/PaymentPage.test.jsx.snap
abutterworth: I think they do no harm, I can checkout the function version, but if it's syntactically messy I'd prefer we just accept these spans.

src/payment/__snapshots__/PaymentPage.test.jsx.snap
abutterworth: @albemarle I found the reason for this, and since fixed it. Our mock stores had coupon data in two places and they didn't match.

src/payment/data/sagas.js
abutterworth: 👍

src/payment/data/sagas.js
abutterworth: We have been using a pattern of naming where UI focused work is named after what's happening and functions that make api calls are named with HTTP verbs.

PR #135, opened 2019-07-28
fix: do not lose decimals when parsing numbers

abutterworth:
Swap `Number.parseInt(results.orderTotal, 10)` for `Number(results.orderTotal)`

Fixes ARCH-1064


abutterworth:
> I assume these are the only occurrences of `parseInt`?

There's in credit card number form, but it's not a problem there.

PR #138, opened 2019-07-28
fix: update discount line to properly display numbers

abutterworth:
Update discounts to display properly given recent backend changes.
Screen Shot 2019-07-28 at 5 08 29 PM


src/payment/OrderSummary.jsx
```diff
@@ -109,13 +109,22 @@ function DiscountOffers({ offers, discounts }) {
...
           <FormattedMessage
             id="payment.summary.discount.offer"
-            defaultMessage="{benefitValue} discount provided by {provider}."
+            defaultMessage="{benefit} discount provided by {provider}."
             description="A description of a discount offer applied to a basket."
-            values={{ benefitValue, provider }}
+            values={{
+              benefit: (
+                <FormattedNumber
+                  value={benefitValue / 100}
```
albemarle: This seems like it would only work for percentage?  Or am I misreading it?

package.json
```diff
@@ -36,7 +36,7 @@
     "@edx/frontend-component-site-header": "2.3.0",
     "@edx/frontend-i18n": "2.1.0",
     "@edx/frontend-logging": "2.1.0",
-    "@edx/paragon": "^6.2.0",
```
davidjoy: I'm not sure how this change (from Renovate) ended up in here.  Sorta concerning.

src/payment/OrderSummary.jsx
abutterworth: Good catch – David fixed it

PR #141, opened 2019-07-29
fix: handle failure to get access token

abutterworth:
ARCH-1066: Sometimes we fail to get an access token. Handle this case by showing our error page and add logging for future investigation.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -20,6 +20,7 @@ import { configureApiService as configurePaymentApiService } from './payment';
 
 import './index.scss';
 import App from './components/App';
+import ErrorPage from './common/components/ErrorPage';
 
 if (configuration.ENVIRONMENT !== 'production') {
   import(/* webpackChunkName: "react-axe" */ 'react-axe')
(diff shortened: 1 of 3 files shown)


src/index.jsx
```diff
@@ -73,13 +74,17 @@ apiClient.ensurePublicOrAuthenticationAndCookies(
...
     const { store, history } = configure();
 
-    ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
-
     if (accessToken) {
+      ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
       identifyAuthenticatedUser(accessToken.userId);
+      sendPageEvent();
     } else {
+      // This should never happen, but it does sometimes.
```
albemarle: :D

src/index.jsx
```diff
@@ -73,13 +74,17 @@ apiClient.ensurePublicOrAuthenticationAndCookies(
...
+      // Add logging to learn more.
+      ReactDOM.render(<ErrorPage />, document.getElementById('root'));
+      NewRelicLoggingService.logError('Empty accessToken returned from ' +
+        'ensurePublicOrAuthenticationAndCookies callback.');
```
albemarle: Is there anything else useful we can add about the state?

src/index.jsx
```diff
@@ -73,13 +74,17 @@ apiClient.ensurePublicOrAuthenticationAndCookies(
...
-    ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
-
     if (accessToken) {
+      ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
       identifyAuthenticatedUser(accessToken.userId);
+      sendPageEvent();
     } else {
+      // This should never happen, but it does sometimes.
+      // Add logging to learn more.
+      ReactDOM.render(<ErrorPage />, document.getElementById('root'));
```
robrap: Is there a way to just use ErrorBoundary?

abutterworth: We're handling this error outside the scope of React, so not directly no.

src/index.jsx
abutterworth: @robrap Is there anything useful here? I can't think of anything.

robrap: @albemarle: FYI: We've got other related frontend-auth code changes to help log and track this down.
@abutterworth: We should think about the right interface for ensurePublicOrAuthenticationAndCookies though.  It seems wrong to be calling this callback in this error condition.

PR #143, opened 2019-07-29
feature(prices): move localized currency disclaimer closer to prices

albemarle:
We thought the message about currency conversions being estimates would make more sense closer to the actual prices.

Also italicized the message to match the legacy page.

moved disclaimer higher

### DIFF
--- a/src/payment/order-summary/OrderSummary.jsx
+++ b/src/payment/order-summary/OrderSummary.jsx
@@ -4,12 +4,13 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { FormattedMessage } from '@edx/frontend-i18n';
 
-import { basketSelector } from '../data/selectors';
+import { orderSummarySelector } from '../data/selectors';
 import BulkOrderSummaryTable from './BulkOrderSummaryTable';
 import SummaryTable from './SummaryTable';
 import TotalTable from './TotalTable';
 import CouponForm from './CouponForm';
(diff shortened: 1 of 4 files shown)


src/payment/order-summary/OrderSummary.jsx
```diff
@@ -4,12 +4,13 @@ import PropTypes from 'prop-types';
...
 import { FormattedMessage } from '@edx/frontend-i18n';
 
-import { basketSelector } from '../data/selectors';
+import { orderSummarySelector } from '../data/selectors';
 import BulkOrderSummaryTable from './BulkOrderSummaryTable';
 import SummaryTable from './SummaryTable';
 import TotalTable from './TotalTable';
 import CouponForm from './CouponForm';
 import DiscountOffers from './DiscountOffers';
+import CurrencyDisclaimer from '../CurrencyDisclaimer';
```
davidjoy: Nit: Could move CurrencyDisclaimer into the order-summary directory.

albemarle: I considered that, if we assume it's always going to go with the order summary?

davidjoy: Seems like order summary is wrapping a few disparate things already, so I'd say "suuure!"  Easy to change later if it gets put somewhere else.

albemarle: Did the thing.

PR #154, opened 2019-07-30
fix: disable buttons while basket is processing

abutterworth:
ARCH-1068: During any api request to basket endpoints disable any buttons that would modify the basket in any way. This ensures the frontend app will contain basket data that matches the server's

### DIFF
--- a/src/payment/PaymentMethodSelect.jsx
+++ b/src/payment/PaymentMethodSelect.jsx
@@ -8,7 +8,9 @@ import AcceptedCardLogos from './assets/accepted-card-logos.png';
 import { ApplePayButtonContainer } from './apple-pay';
 import { PayPalButton } from './paypal';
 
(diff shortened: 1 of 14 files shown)


src/payment/PaymentMethodSelect.jsx
```diff
@@ -8,7 +8,9 @@ import AcceptedCardLogos from './assets/accepted-card-logos.png';
 import { ApplePayButtonContainer } from './apple-pay';
 import { PayPalButton } from './paypal';
 
-function PaymentMethodSelect({ intl, loading }) {
+function PaymentMethodSelect({ intl, loading, isBasketProcessing }) {
+  const submissionDisabled = loading || isBasketProcessing;
```
albemarle: This reads a little odd to me, maybe submitDisabled would do?  Not a big deal though.

PR #157, opened 2019-07-30
Throw an error if application state is empty object.

davidjoy:
**NOTE**: This could use an extra-thoughtful review, as there are currently no tests for this part of our code (and writing them would be difficult).  

Do this by wrapping app initialization in a try/catch and going to error page from there.  Fixes this issue and also catches any other weird errors that may happen.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -2,7 +2,6 @@ import 'babel-polyfill';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {
-  identifyAnonymousUser,
   identifyAuthenticatedUser,
   sendPageEvent,
   configureAnalytics,
@@ -22,30 +21,18 @@ import './index.scss';
 import App from './components/App';
 import ErrorPage from './common/components/ErrorPage';
 
-if (configuration.ENVIRONMENT !== 'production') {
-  import(/* webpackChunkName: "react-axe" */ 'react-axe')
-    .then(({ default: axe }) => axe(React, ReactDOM, 1000));
-}
-
-const apiClient = getAuthenticatedAPIClient({
-  appBaseUrl: configuration.BASE_URL,
-  authBaseUrl: configuration.LMS_BASE_URL,
-  loginUrl: configuration.LOGIN_URL,
-  logoutUrl: configuration.LOGOUT_URL,
-  csrfTokenApiPath: configuration.CSRF_TOKEN_API_PATH,
-  refreshAccessTokenEndpoint: configuration.REFRESH_ACCESS_TOKEN_ENDPOINT,
-  accessTokenCookieName: configuration.ACCESS_TOKEN_COOKIE_NAME,
-  userInfoCookieName: configuration.USER_INFO_COOKIE_NAME,
-  csrfCookieName: configuration.CSRF_COOKIE_NAME,
-  loggingService: NewRelicLoggingService,
-});
+let apiClient = null;
 
 /**
  * We need to merge the application configuration with the authentication state
  * so that we can hand it all to the redux store's initializer.
  */
 function createInitialState() {
-  return Object.assign({}, { configuration }, apiClient.getAuthenticationState());
+  const authenticationState = apiClient.getAuthenticationState();
+  if (Object.keys(authenticationState).length === 0) {
+    throw new Error('Empty authentication state returned from gettAuthenticationState()');
+  }
+  return Object.assign({}, { configuration }, authenticationState);
 }
 
 function configure() {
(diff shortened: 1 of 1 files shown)


src/index.jsx
```diff
@@ -2,7 +2,6 @@ import 'babel-polyfill';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {
-  identifyAnonymousUser,
```
davidjoy: @robrap indicated we shouldn't be publishing an identifyAnonymousUser event if we have no accessToken.  So I've removed it and where it was called below.

src/index.jsx
```diff
@@ -22,30 +21,18 @@ import './index.scss';
...
-  loginUrl: configuration.LOGIN_URL,
-  logoutUrl: configuration.LOGOUT_URL,
-  csrfTokenApiPath: configuration.CSRF_TOKEN_API_PATH,
-  refreshAccessTokenEndpoint: configuration.REFRESH_ACCESS_TOKEN_ENDPOINT,
-  accessTokenCookieName: configuration.ACCESS_TOKEN_COOKIE_NAME,
-  userInfoCookieName: configuration.USER_INFO_COOKIE_NAME,
-  csrfCookieName: configuration.CSRF_COOKIE_NAME,
-  loggingService: NewRelicLoggingService,
-});
+let apiClient = null;
```
davidjoy: We need the apiClient all over in this file, so it needs to be scoped to the top.  (i.e., it can't be in a try/catch)

src/index.jsx
```diff
@@ -22,30 +21,18 @@ import './index.scss';
...
 
 /**
  * We need to merge the application configuration with the authentication state
  * so that we can hand it all to the redux store's initializer.
  */
 function createInitialState() {
-  return Object.assign({}, { configuration }, apiClient.getAuthenticationState());
+  const authenticationState = apiClient.getAuthenticationState();
+  if (Object.keys(authenticationState).length === 0) {
```
davidjoy: This is the change.  If authenticationState, which is an object, has no keys in Object.keys, then that's our error condition.

src/index.jsx
```diff
@@ -22,30 +21,18 @@ import './index.scss';
...
+    throw new Error('Empty authentication state returned from gettAuthenticationState()');
```
davidjoy: This will bubble up to our try/catch below.

src/index.jsx
```diff
@@ -69,22 +56,41 @@ function configure() {
   };
 }
 
-apiClient.ensurePublicOrAuthenticationAndCookies(
-  window.location.pathname,
-  (accessToken) => {
-    const { store, history } = configure();
+function initialize(accessToken) {
```
davidjoy: This function is just a named version of the callback/closure we used to give to `ensurePublicOrAuthenticationAndCookies`.

src/index.jsx
```diff
@@ -69,22 +56,41 @@ function configure() {
...
+  const { store, history } = configure();
+
+  if (accessToken) {
+    ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
+    identifyAuthenticatedUser(accessToken.userId);
+    sendPageEvent();
+  } else {
+    // This should never happen, but it does sometimes.
```
davidjoy: This section has changed.  Now we throw an error which gets caught and handled below.  We used to render the error page and log to new relic directly in here.

src/index.jsx
```diff
@@ -69,22 +56,41 @@ function configure() {
...
-      ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
-      identifyAuthenticatedUser(accessToken.userId);
-      sendPageEvent();
-    } else {
-      // This should never happen, but it does sometimes.
-      // Add logging to learn more.
-      ReactDOM.render(<ErrorPage />, document.getElementById('root'));
-      NewRelicLoggingService.logError('Empty accessToken returned from ' +
-        'ensurePublicOrAuthenticationAndCookies callback.');
-      identifyAnonymousUser();
```
davidjoy: This line to identify an anonymous user is intentionally removed.  We don't want to do that here - this is an error condition, not a normal thing for anonymous users.

src/index.jsx
```diff
@@ -69,22 +56,41 @@ function configure() {
...
+    accessTokenCookieName: configuration.ACCESS_TOKEN_COOKIE_NAME,
+    userInfoCookieName: configuration.USER_INFO_COOKIE_NAME,
+    csrfCookieName: configuration.CSRF_COOKIE_NAME,
+    loggingService: NewRelicLoggingService,
+  });
+
+  apiClient.ensurePublicOrAuthenticationAndCookies(window.location.pathname, initialize);
+} catch (e) {
+  ReactDOM.render(<ErrorPage />, document.getElementById('root'));
+  NewRelicLoggingService.logError(e.message);
```
davidjoy: Note that we're using NewRelicLoggingService directly here (which I didn't change).  In the long term, we might need to think about how we use frontend-logging's interface, rather than calling the NewRelic service directly.

This may be complicated by the fact that the logging service needs to be configured _in_ the code we're wrapping with this try/catch.

src/index.jsx
robrap: Do we know it will be an object, or should we check first?

PR #179, opened 2019-07-31
fix: log error on potential redirect loop

abutterworth:
Inspired by ARCH-1074. If the payment page is directed by the api to redirect to the url it is already on, log an error.

### DIFF
--- a/src/payment/data/service.js
+++ b/src/payment/data/service.js
@@ -1,4 +1,5 @@
 import pick from 'lodash.pick';
+import { logError } from '@edx/frontend-logging';
 
 import { configureApiService as configureCybersourceApiService } from '../cybersource';
 import { configureApiService as configurePayPalApiService } from '../paypal';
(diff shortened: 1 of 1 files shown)


src/payment/data/service.js
```diff
@@ -32,6 +33,13 @@ export function configureApiService(newConfig, newApiClient) {
   apiClient.interceptors.response.use((response) => {
     const { status, data } = response;
     if (status >= 200 && status < 300 && data && data.redirect) {
+      // Redirecting this SPA to itself is likely to cause
+      // a redirect loop.
+      if (window.location.href === data.redirect) {
```
albemarle: Do we want global instead?

albemarle:
I wonder if we should also compare them excluding query params?

src/payment/data/service.js
abutterworth: I think* it should not be an issue even in a gatsby world since this will never fire until runtime. I'll update anyway.

src/payment/data/service.js
```diff
@@ -32,6 +33,13 @@ export function configureApiService(newConfig, newApiClient) {
...
+        logError('The api response is directing the basket page to redirect to itself', {
```
robrap: Maybe: `'The api response is redirecting to the same payment page url'`?

albemarle: "Houston, we have a loop."

PR #181, opened 2019-07-31
Improve info logging custom metrics and log unknown errors as errors

davidjoy:


src/common/serviceUtils.js
```diff
@@ -87,25 +86,32 @@ function handleApiMessages(messages) {
...
   if (error.response && error.response.data.messages !== undefined) {
-    logInfo('API Messages', error.response.data.messages);
+    logInfo('API Messages', { messages: error.response.data.messages });
     handleApiMessages(error.response.data.messages);
   }
 
   // Single API error
   if (error.response && error.response.data.error_code) {
-    logInfo('API Error', error.response.data.errors);
+    logInfo('API Error', {
```
davidjoy: This was just plain wrong, far as I can tell.  If it has error_code it needs to parse that apart, not look at errors.

src/common/serviceUtils.js
```diff
@@ -87,25 +86,32 @@ function handleApiMessages(messages) {
...
   if (error.response && error.response.data.errors !== undefined) {
-    logInfo('API Errors', error.response.data.errors);
+    logInfo('API Errors', { errors: error.response.data.errors });
     handleApiErrors(error.response.data.errors);
   }
 
   // API messages
   if (error.response && error.response.data.messages !== undefined) {
-    logInfo('API Messages', error.response.data.messages);
+    logInfo('API Messages', { messages: error.response.data.messages });
```
davidjoy: Left the message generic and added the custom metric name.  It's pretty effective to filter by appName in my limited experience.

src/common/serviceUtils.js
```diff
@@ -118,3 +124,21 @@ export function handleRequestError(error) {
   logAPIErrorResponse(error);
   throw error;
 }
+
+export function isUnknownError(error) {
```
davidjoy: Helps us know which errors are unknown in the sagas, so we can log them as errors (and not just log them to info above)

src/feedback/data/sagas.js
```diff
@@ -3,7 +3,7 @@ import { put } from 'redux-saga/effects';
 import { addMessage, clearMessages } from './actions';
 import { MESSAGE_TYPES } from './constants';
 
-export function* handleErrors(e, clearExistingMessages) {
+export function* handleErrorFeedback(e, clearExistingMessages) {
```
davidjoy: I changed the name of these two functions to indicate they're not just doing all our error handling.  They're doing something much more specific that doesn't involve logging anything!

src/payment/cybersource/data/sagas.js
```diff
@@ -49,9 +52,12 @@ export function* handleSubmitPaymentCybersource(action) {
       card_expiry_date: [cardExpirationMonth.padStart(2, '0'), cardExpirationYear].join('-'),
     };
     generateAndSubmitForm(configuration.CYBERSOURCE_URL, cybersourcePaymentParams);
-  } catch (e) {
+  } catch (error) {
     yield put(submitPaymentCybersourceFailure());
-    yield call(handleErrors, e, true);
+    yield call(handleErrorFeedback, error, true);
+    if (isUnknownError(error)) {
```
davidjoy: I added this `if` to our various catches that weren't actually logging errors.

src/common/serviceUtils.js
robrap: I'd really like this message to be more clear, even if it doesn't include the app name.  Something like:
```
API error with expected user messages. See messages metric for details.
```
I'm writing this from the experience of looking at "API Messages" in insights and having no clue what it meant.  I don't see why we wouldn't want the message to make things more immediately clear.

src/common/serviceUtils.js
```diff
@@ -87,25 +86,32 @@ function handleApiMessages(messages) {
 export function handleRequestError(error) {
   // Validation errors
   if (error.response && error.response.data.field_errors) {
-    logInfo('Field Errors', error.response.data.field_errors);
+    logInfo('API field errors returned to client in field_errors field', { fieldErrors: error.response.data.field_errors });
     handleFieldErrors(error.response.data.field_errors);
```
abutterworth: Note for future refactor: this series of handleXXXXXErrors that throw and error may be better named as `throwFieldError` or something so it's clearer that we won't be logging the default case at the end.

src/common/serviceUtils.js
```diff
@@ -87,25 +86,32 @@ function handleApiMessages(messages) {
...
-    logInfo('API Messages', error.response.data.messages);
+    logInfo('API errors returned to client in messages field', { messages: error.response.data.messages });
     handleApiMessages(error.response.data.messages);
   }
 
   // Single API error
   if (error.response && error.response.data.error_code) {
-    logInfo('API Error', error.response.data.errors);
+    logInfo('API error returned to client as response', {
+      errors: [
```
robrap: 1. Curious why this is an array?
2. We should loop back to this later.  `logAPIErrorResponse` has a lot of axios response parsing logic that should be factored out so it can be used with logInfo as well, and not have to be re-written.  See https://github.com/edx/frontend-logging/blob/master/src/NewRelicLoggingService.js#L50-L58

src/common/serviceUtils.js
```diff
@@ -118,3 +124,21 @@ export function handleRequestError(error) {
   logAPIErrorResponse(error);
   throw error;
 }
+
+export function isUnknownError(error) {
+  if (error.response) {
+    if (error.response.data.field_errors !== undefined) {
```
albemarle: Tiny: would it be nicer to do ```const {field_errors, errors, messages, error_code} = error.response.data;``` ?

albemarle: (Also, does data always exist for sure?)

src/feedback/data/sagas.js
albemarle: showErrorFeedback?  showErrorsToUser?  Hmm.

src/common/serviceUtils.js
```diff
@@ -87,25 +86,32 @@ function handleApiMessages(messages) {
...
   }
 
   // Single API error
   if (error.response && error.response.data.error_code) {
-    logInfo('API Error', error.response.data.errors);
+    logInfo('API error returned to client as response', {
+      errors: [
+        {
+          error_code: error.response.data.error_code,
+          user_message: error.response.data.user_message,
```
robrap: Could this break things if there is no `error.response.data.user_message`?  My main concern about this PR is that we don't introduce errors.  We can spend time testing that it does the right thing tomorrow.

robrap: Actually, I’m not clear on what case this is and why we aren’t using `logAPIErrorResponse(error);` here. You can fill me in tomorrow.

src/common/serviceUtils.js
davidjoy: Agreed - I totally got confused by that when I was reading it.

src/common/serviceUtils.js
davidjoy: For 1. - Not for a particularly good reason.  We were massaging a single error into an array elsewhere so I think I had it on the brain.
2. - Sounds great.  In working on this I felt like there were too many layers, so it'd be nice if all collapsed down a bit.

src/common/serviceUtils.js
davidjoy: Yeeessssss

src/common/serviceUtils.js
davidjoy: Not really, in my opinion.  If user_message isn't there it'll just be undefined.  If _data_ isn't there, we've got bigger problems, but I don't think that's possible.

robrap: I see regarding user_message. Thanks.
I need to see the code on my computer tomorrow. I’m unclear what this situation is, and the message didn’t clarify.

src/common/serviceUtils.js
robrap: I think I'm now more clear on these different cases.  I think, but correct me if I am wrong, that this might be leftover from a different/earlier api response design where `error_code` was put at the top-level of the response?  If you agree, can we use logging (potentially even logError) to help us confirm that this is never happening so we can loop back and delete this case altogether?

src/common/serviceUtils.js
robrap: I've got a slightly different design in mind, but I think we can land this and loop back.  Happy to discuss my thoughts as well, but quicker in person.

davidjoy:
This PR didn't really do quite what we thought it did.  Closing without merging.

PR #185, opened 2019-08-02
If the basket response indiciates we should redirect, show a spinner.

davidjoy:
I'd like to deploy this to stage so we can test it more fully - I wasn't able to get an actual redirect back from the server while developing.

### DIFF
--- a/src/payment/PaymentPage.jsx
+++ b/src/payment/PaymentPage.jsx
@@ -30,10 +30,7 @@ class PaymentPage extends React.Component {
   }
 
   renderEmptyMessage() {
-    const {
-      dashboardURL,
-      supportURL,
-    } = this.props;
+    const { dashboardURL, supportURL } = this.props;
 
     return (
       <div className="card">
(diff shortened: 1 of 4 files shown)


src/payment/PaymentPage.jsx
```diff
@@ -169,8 +171,9 @@ class PaymentPage extends React.Component {
             ),
           }}
         />
-        {loaded && isEmpty ? this.renderEmptyMessage() : null}
-        {loading || !isEmpty ? this.renderBasket() : null}
+        {(!loaded || loading || isValidBasket) ? this.renderBasket() : null}
```
albemarle: What do you think of making a descriptively-named variable for the conditional part of this?  On the face of it, its not clear reading this why if it's loading we render the basket, but if it's a redirect we render loading.

davidjoy: Admittedly the names are getting a bit fuzzy.  I agree.  🤔

davidjoy: I tried to clean it up!

src/payment/PaymentPage.jsx
```diff
@@ -30,10 +30,7 @@ class PaymentPage extends React.Component {
   }
 
   renderEmptyMessage() {
-    const {
-      dashboardURL,
-      supportURL,
-    } = this.props;
+    const { dashboardURL, supportURL } = this.props;
```
davidjoy: Formatted the file which fixed a bunch of indentation and stuff that I hadn't even noticed.  Figured I'd just keep it since I was going back and trying to make things clearer anyway. :)

albemarle: Are these things that are worth putting into our eslint config?

src/payment/PaymentPage.jsx
```diff
@@ -51,44 +48,50 @@ class PaymentPage extends React.Component {
...
     return (
       <PageLoading srMessage={this.props.intl.formatMessage(messages['payment.loading.payment'])} />
     );
   }
 
+  /**
+   * We show the basket view for all cases except: 1) an empty basket, and 2) when we're going to
+   * perform a redirect.  That means that sometimes it's used during loading, in which case it shows
+   * a "skeleton" view of the left-hand side of the interface until the actual content arrives.
+   */
```
albemarle: I think this is enough as is, but it did make me think, maybe SkeletonView (LoadingView?) should be its own thing?  It's fine either way though.

src/payment/PaymentPage.jsx
```diff
@@ -126,36 +129,47 @@ class PaymentPage extends React.Component {
           aria-label={this.props.intl.formatMessage(messages['payment.section.payment.details.label'])}
           className="col-md-7 pl-md-5"
         >
-          {isFreeBasket ? <PlaceOrderButton /> : (
+          {isFreeBasket ? (
```
albemarle: ♫  And this basket you cannot change... ♫

src/payment/PaymentPage.jsx
```diff
@@ -126,36 +129,47 @@ class PaymentPage extends React.Component {
...
               <PaymentForm />
             </React.Fragment>
-        )}
+          )}
         </section>
       </div>
     );
   }
 
+  renderContent() {
```
albemarle: Oh, this is much clearer!

src/payment/PaymentPage.jsx
davidjoy: Hm, good point - I think my formatter (Prettier) may be taking some extra liberties beyond what our eslint config says.

src/payment/PaymentPage.jsx
davidjoy: Thought about that - the problem/reason I didn't change it is that the left-hand side may be a skeleton, but the right-hand side of renderBasket() is all the real content.  I think it'd almost be just as confusing if renderSkeleton() and renderBasket() were both rendering the real forms on the right-hand side. :)  Either way the method name isn't quite describing what's going on in it, imho...

src/payment/PaymentPage.jsx
```diff
@@ -126,36 +129,47 @@ class PaymentPage extends React.Component {
...
+    const { isEmpty, isRedirect } = this.props;
+
+    if (isEmpty) {
+      return this.renderEmptyMessage();
+    }
+
+    if (isRedirect) {
```
robrap: Nit: I know you have `!basket.redirect` in `isEmpty`, but it would feel safer(clearer?) to have the isRedirect if statement first.

davidjoy: 👍Good nit.  Fixing.

PR #191, opened 2019-08-05
Remove codecov and coveralls from Travis config.

davidjoy:
We aren’t using coveralls for this repo, and we want to let Github manage our usage of codecov.  Currently our codecov config is failing our builds in Travis, causing the transifex jenkins jobs to fail downstream.  This PR should clear this up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,3 @@ script:
   - npm run test
   - npm run build
   - npm run is-es5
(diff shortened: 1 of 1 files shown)


davidjoy:
This PR won't work - we need the codecov line in Travis to report test coverage data to codecov so it can be shown in Github.  They work together. Closing.

PR #196, opened 2019-08-06
docs: HOWTO for localized currency

albemarle:


docs/how_tos/localized_currency.rst
```diff
@@ -0,0 +1,66 @@
...
+
+This HOWTO explains what the localized currency cookie is, how it gets set, and how it is connected to our LocalizedPrice component.
+
+.. contents:: Table of Contents
+
+***************************
+What is localized currency?
+***************************
+
+This is when we show users prices on the checkout page in what we think is the correct currency for where they are. These prices can be rounded, since the final checkout charge will still be in U.S. dollars (USD).
```
robrap: Can we use ‘payment page’ instead?

docs/how_tos/localized_currency.rst
```diff
@@ -0,0 +1,66 @@
...
+
+***********************
+When is the cookie set?
+***********************
+
+The ``edx-price-l10n`` cookie is set when a user visits a course's About page. (Since August 2019, the cookie is set whether the About page is generated by Drupal or Prospectus.)  This cookie is set based on the user's location as per the browser's `geolocation API <https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API>`_.
```
robrap: We need a way to properly document this. It is important, and this feels like the right place, but it refers to edx private repos in a public Open edX repo.

Maybe separate the cookie contract details from a “For example on edx.org...” section? Also, it might be better to point to a doc in Prospectus so this can be updated along with the repo as things change?

rgraber: Happened to see this, if it matters the cookie is now also set on all program about pages as well. Also we switched to using IP location from cloudflare.

albemarle: Good info, thanks!

albemarle: @robrap You're right that the documentation on where the cookie is set should live in that repo.  (But it's actually two repos, now.)

To my knowledge there is no other doc to point to.  How about if I replace this section with "...is set when a user visits certain pages (check with edX for details)" and delete the details about geolocation.  Then this doc would include only the cookie contract, not the details about how and when edX code sets it.

robrap: Maybe just state that this code assumes the cookie is set before this application is loaded. I recommend dropping the refernce “...edX for details”, because I don’t think it adds much here. Thank you!

PR #197, opened 2019-08-07
Cease to use angle brackets in messages.

davidjoy:
They break transifex.


albemarle:
After tonight's push job, Transifex should invalidate all translations for those keys, and put the new English text up as untranslated again.

PR #201, opened 2019-08-07
fix: show spinner when loading a coupon redeem basket

abutterworth:
ARCH-1101: Do not show the loading skeleton when loading the page with a get parameter for indicating this is a coupon redeem basket likely to redirect


src/payment/data/selectors.js
```diff
@@ -17,6 +17,8 @@ export const localizedCurrencySelector = (state) => {
 
 export const basketSelector = state => ({
   ...state[storeName].basket,
+  isCouponRedeemRedirect: state.queryParameters &&
+    state.queryParameters.coupon_redeem_redirect == 1, // eslint-disable-line eqeqeq
```
albemarle: Would be nice to have a util function like getQueryParamValue.

src/common/getQueryParameters.js
```diff
@@ -0,0 +1,13 @@
+export default function getQueryParameters(search = window.location.search) {
+  const keyValueFragments = search
+    .slice(search.indexOf('?') + 1)
+    .split('&')
+    .filter(hash => hash !== '');
```
albemarle: Let's add some unit tests around this.

src/common/getQueryParameters.js
```diff
@@ -0,0 +1,13 @@
+export default function getQueryParameters(search = global.location.search) {
+  const keyValueFragments = search
+    .slice(search.indexOf('?') + 1)
+    .split('&')
```
davidjoy: Unfortunately this will break if there's an &amp; or any other URL encoded entity in the query string... 

Wondering if we can use one of the two approaches described here? https://davidwalsh.name/query-string-javascript

davidjoy: (I know that the particular use case we have for this function won't care about the above, but if we're going to have a general-purpose function, then we should probably consider it)

src/index.jsx
```diff
@@ -32,7 +33,8 @@ function createInitialState() {
   if (Object.keys(authenticationState).length === 0) {
     throw new Error('Empty authentication state returned from gettAuthenticationState()');
   }
-  return Object.assign({}, { configuration }, authenticationState);
+  const queryParameters = getQueryParameters();
```
davidjoy: Again, probably not a problem for us right now, but downstream we should be aware that this won't actually update redux when the query params change.

davidjoy:
Up for debate whether the above should be something we change... Happy to talk about it.

PR #202, opened 2019-08-07
Load optimizely last.

davidjoy:
Dropped html-webpack-new-relic-plugin dependency so that we could keep the new relic script above our own JS scripts.

A future version of html-webpack-plugin will allow us to inject the scripts with `htmlWebpackPlugin.tags.bodyTags` instead of our manual looping.


webpack/webpack.prod.config.js
```diff
@@ -165,9 +164,11 @@ module.exports = Merge.smart(commonConfig, {
     }),
     // Generates an HTML file in the output directory.
     new HtmlWebpackPlugin({
-      inject: true, // Appends script tags linking to the webpack bundles at the end of the body
+      inject: false, // Manually inject head and body tags in the template itself.
       template: path.resolve(__dirname, '../public/index.html'),
       optimizelyId: process.env.OPTIMIZELY_PROJECT_ID,
+      newRelicLicenseKey: process.env.NEW_RELIC_LICENSE_KEY || 'fake_app',
```
robrap: Reverse the fakes and good to go. :)

PR #205, opened 2019-08-08
Disable the Apple Pay button after initial click.

davidjoy:
And re-enable it once done (cancel, fail, success)

### DIFF
--- a/src/payment/PaymentPage.jsx
+++ b/src/payment/PaymentPage.jsx
@@ -7,7 +7,7 @@ import { Hyperlink } from '@edx/paragon';
 import messages from './PaymentPage.messages';
 
 // Actions
-import { fetchBasket } from './data/actions';
+import { fetchBasket, submitPayment } from './data/actions';
 
 // Selectors
 import { paymentSelector } from './data/selectors';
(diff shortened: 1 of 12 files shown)


src/payment/data/reducers.js
```diff
@@ -24,6 +25,12 @@ const basket = (state = basketInitialState, action = null) => {
   switch (action.type) {
     case BASKET_DATA_RECEIVED: return { ...state, ...action.payload };
 
+    // For submission, we only really need to know whether it's actively submitting or not.
+    case submitPayment.TRIGGER: return { ...state, submitting: true };
+    case submitPayment.SUCCESS: return { ...state, submitting: false };
```
abutterworth: Since is triggered on success or failure, we don't need submitting false for the success of failure actions.

src/payment/PaymentPage.jsx
```diff
@@ -133,7 +137,13 @@ class PaymentPage extends React.Component {
             <PlaceOrderButton />
           ) : (
             <React.Fragment>
-              <PaymentMethodSelect loading={loading} isBasketProcessing={isBasketProcessing} />
+              <PaymentMethodSelect
+                submitPayment={this.props.submitPayment}
+                submitPaymentSuccess={this.props.submitPaymentSuccess}
```
abutterworth: If the ApplePayButtonContainer was not part of the ApplePay "module" we could logically just connect that to these actions and skip passing props down through multiple components.

src/payment/data/reducers.js
davidjoy: Not sure I follow... `TRIGGER` happens at the beginning, `SUCCESS` and `FAILURE` are ending conditions so we need to set `submitting` back to false since we're done.

src/payment/PaymentPage.jsx
davidjoy: Truth

davidjoy:
"1005%" coverage.  Whoops.

PR #206, opened 2019-08-09
refactor: unify handling of payment submissions

abutterworth:
Addresses large portions of ARCH-1072 (Frontend Payments code cleanup).

- Strips payment methods down to simple components and services that include a `checkout(<basket>, <args>)` function that returns a promise.
- All payment submissions are now triggered by the `submitPayment` routine.


src/payment/data/sagas.js
```diff
@@ -88,14 +96,39 @@ export function* handleUpdateQuantity({ payload }) {
...
+    yield put(submitPayment.success(result));
+  } catch (error) {
+    if (error.aborted === true) {
+      // This an a user aborted action, do nothing
+    } else {
+      yield put(submitPayment.failure(error));
+      if (error.code) {
+        // Client side generated errors are simple error objects and
+        // here we wrap them into the same format as the api returns them in
+        yield call(handleErrors, { errors: [error] }, true);
```
davidjoy: One thing you might want to tweak - in this app, we actually _only_ use the `messages` key for server side errors.  Robert and I had talked about removing the handling for `errors` and `field_errors` since it wasn't being used, though we didn't actually bother.  Put another way, this is actually not technically the same format as the APIs. :)

PR #209, opened 2019-08-09
fix: form autocomplete

julianajlk:
Added autocomplete attributes to forms. 
[ARCH-1112](https://openedx.atlassian.net/browse/ARCH-1112).


wittjeff:
Thanks!

PR #214, opened 2019-08-12
refactor: reorganization of components

abutterworth:
Code is now organized broadly into `cart/` and `checkout/`

`Cart.jsx` and `Checkout.jsx` are the new interesting additions here. They are connected components that have enabled the dumbing down or elimination of some others. Checkout contains all `submitPayment` action dispatches and `sendTrackEvent` calls.


src/payment/cart/UpdateQuantityForm.jsx
```diff
@@ -4,8 +4,8 @@ import { connect } from 'react-redux';
 import { FormattedMessage } from '@edx/frontend-i18n';
 import { StatefulButton, Input } from '@edx/paragon';
 
-import { updateQuantity } from './data/actions';
-import { updateQuantityFormSelector } from './data/selectors';
+import { updateQuantity } from '../data/actions';
+import { updateQuantityFormSelector } from '../data/selectors';
```
davidjoy: This is a nit, but is maybe interesting to think about... technically this breaks the 'rules' of importing something from your parent - i.e., taking for granted something outside your module/context.  If I were to try to avoid that, I might pass the action in as a parameter to this component.  The selector is harder, as it depends on selectors in the parent too.  I think we'd have to jump through some hoops to get the selector to be context agnostic, and it's clearly not worth it here.  But, that said, if the selector were in this sub-directory's data dir instead of in the parent, that'd at least be marginally closer.  ¯\_(ツ)_/¯ 

This all has me thinking that there should probably be a difference between a module and a sub-directory merely for organization.  It also reminds me that redux breaks all rules of encapsulation and modularity by forcing components to be aware of the global position of their data. 😡

src/payment/cart/order-details/data/selectors.js
```diff
@@ -1,7 +1,7 @@
 import { createSelector, createStructuredSelector } from 'reselect';
 
-import { getModuleState } from '../../../common/utils';
-import { emailSelector } from '../../../common/selectors';
+import { getModuleState } from '../../../../common/utils';
```
davidjoy: frontend-common can't get here soon enough... 🤦‍♂

src/payment/checkout/Checkout.jsx
```diff
@@ -0,0 +1,176 @@
...
+    // Check for PayPal and Free Basket as well
+    sendTrackEvent(
+      'edx.bi.ecommerce.basket.payment_selected',
+      { type: 'click', category: 'checkout', paymentMethod: 'Apple Pay' },
+    );
+
+    this.props.submitPayment({ method: 'apple-pay' });
+  }
+
+  handleSubmitCybersource = (formData) => {
```
davidjoy: I think this is still miles ahead of where it was, but it _is_ a bit confusing to have `handleSubmitCybersource` and `handleSubmitCybersourceButtonClick` side by side... it's like... why are there two?  Which one is doing what?  

Is it possible to give the payment form a single click handler and call both from it?  Or do they get called at different times? (assuming I'll find out when I read `PaymentForm`...)

src/payment/checkout/FreeCheckoutOrderButton.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from '@edx/frontend-i18n';
+import { Hyperlink } from '@edx/paragon';
+
+function FreeCheckoutOrderButton({ ecommerceURL, onClick }) {
```
davidjoy: Yeeeesssssss.  So much better.

src/payment/checkout/payment-form/CardHolderInformation.jsx
```diff
@@ -37,7 +37,7 @@ export class CardHolderInformationComponent extends React.Component {
   }
 
   render() {
-    const { submitting, showBulkEnrollmentFields } = this.props;
+    const { disabled, showBulkEnrollmentFields } = this.props;
```
davidjoy: Yay for naming consistency!

src/payment/checkout/payment-form/PaymentForm.jsx
```diff
@@ -222,7 +199,7 @@ export class PaymentFormComponent extends React.Component {
                   type="submit"
                   className="btn btn-primary btn-lg btn-block"
                   state={submitButtonState}
-                  onClick={this.handleSubmitButtonClick}
+                  onClick={this.props.onSubmitButtonClick}
```
davidjoy: Hm!  Continuing the thought from above... So the `onSubmitButtonClick` and `onSubmitPayment` handlers do basically fire at the same time... I wonder if we could effectively remove the one on this button and just rely on the form `onSubmit`?  On the outside, the handler would just do both things.  Might be easier to understand for others.

src/payment/cart/UpdateQuantityForm.jsx
abutterworth: Hah yea, I do think it would be good to make an obvious distinct when a folder house s a module vs just organizes some code.

src/payment/cart/order-details/data/selectors.js
abutterworth: We could make an alias in webpack and clean this up but may be not worth it at the moment

src/payment/checkout/Checkout.jsx
abutterworth: The comment below `handleSubmitCybersource` and above `handleSubmitCybersourceButtonClick` says:
```
  // The payment form does client side validation that happens before
  // the submit handler above is fired. We send the tracking event here
  // on click of the submit button for parity with the old page.
```
Maybe there's a better way to describe the reason?

PR #220, opened 2019-08-14
Fixes a variety of warnings in tests and the running app.

davidjoy:
The only warnings left are about ReactTooltip and StatusAlert (paragon) using lifecycle methods that are deprecated.

- Fixes babel-polyfill warning in dev
- Fixes warnings about using navigation in cybersource and apple-pay service tests.
- Fixes warnings about required props for PaymentForm
- Fixes warnings about null font awesome icons.
- Fixes warnings about null title and imageUrl in ProductLineItem, which happens frequently in non-production environments.
- Fixes **test failure** in AlertList which was passing a function for FallbackErrorMessage instead of a component.  I'm not 100% on why this suddenly started failing.
- Updates dependencies to remove lifecycle warning messages: react-redux, react-router, react-router-dom, and redux-form.


webpack/webpack.dev.config.js
```diff
@@ -116,9 +116,11 @@ module.exports = Merge.smart(commonConfig, {
     }),
     // Generates an HTML file in the output directory.
     new HtmlWebpackPlugin({
-      inject: true, // Appends script tags linking to the webpack bundles at the end of the body
+      inject: false, // Appends script tags linking to the webpack bundles at the end of the body
```
davidjoy: This being true here was causing us to double-add our code in development. 😲

PR #225, opened 2019-08-15
fix: update is basket processing state

abutterworth:
Fixes doubling clicking errors described in ARCH-1109. Coupon and UpdateQuantity reducers now set isBasketProcessing directly rather than relying on derived state. Leverages the REQUEST action in redux saga routines to toggle isBasketProcessing. Sagas will noop if they are triggered when isBasketProcessing

### DIFF
--- a/src/payment/data/sagas.js
+++ b/src/payment/data/sagas.js
@@ -24,6 +24,11 @@ export const paymentMethods = {
   'apple-pay': checkoutApplePay,
 };
 
+
+function* isBasketProcessing() {
+  return yield select(state => state.payment.basket.isBasketProcessing);
+}
+
 export function* handleFetchBasket() {
   try {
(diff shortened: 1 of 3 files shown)


src/payment/data/sagas.js
```diff
@@ -43,6 +48,8 @@ export function* handleFetchBasket() {
 }
 
 export function* handleAddCoupon({ payload }) {
+  if (yield isBasketProcessing()) return;
```
nasthagiri: Future: We may want to DRY the 3 very similar handlers here since they all have the same structure (by possibly having a helper function that takes in functions as arguments).  Otherwise, it may be prone to future maintenance issues by requiring all handlers to copy-paste calls to `isBasketProcessing`.

abutterworth: 👍

PR #227, opened 2019-08-16
fix: add more error logging for checkout services

abutterworth:
Related to ARCH-1130.

### DIFF
--- a/src/payment/payment-methods/cybersource/service.js
+++ b/src/payment/payment-methods/cybersource/service.js
@@ -54,8 +54,8 @@ export async function checkout(basket, { cardHolderInfo, cardDetails }) {
   );
 
   if (sdnCheckResponse.hits > 0) {
+    /* istanbul ignore next */
     if (config.ENVIRONMENT !== 'test') {
-      /* istanbul ignore next */
       global.location.href = `${config.ECOMMERCE_BASE_URL}/payment/sdn/failure/`;
     }
     throw new Error('This card holder did not pass the SDN check.');
--- a/src/payment/payment-methods/cybersource/service.js
+++ b/src/payment/payment-methods/cybersource/service.js
@@ -1,5 +1,6 @@
 import formurlencoded from 'form-urlencoded';
 import pick from 'lodash.pick';
+import { logAPIErrorResponse } from '@edx/frontend-logging';
 
 import { applyConfiguration } from '../../../common/serviceUtils';
 import { generateAndSubmitForm } from '../../../common/utils';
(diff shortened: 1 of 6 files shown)


src/payment/payment-methods/cybersource/service.js
```diff
@@ -69,7 +81,16 @@ export async function checkout(basket, { cardHolderInfo, cardDetails }) {
         'Content-Type': 'application/x-www-form-urlencoded',
       },
     },
-  );
+  ).catch((error) => {
+    logAPIErrorResponse(error, {
+      messagePrefix: 'Cybersource Submit Error',
+      paymentMethod: 'Cybersource',
```
robrap: Drying this in the future?  Maybe a helper like `getCybersourceErrorCustomAttributes(basketId, paymentErrorType)`.   It could always include `paymentMethod: 'Cybersource'`.  Or, if it is more general and takes `paymentMethod` as a param, this should take a constant for consistency.

src/payment/payment-methods/cybersource/service.js
```diff
@@ -19,36 +20,47 @@ export function configureApiService(newConfig, newApiClient) {
...
   );
 
   if (sdnCheckResponse.hits > 0) {
+    /* istanbul ignore next */
     if (config.ENVIRONMENT !== 'test') {
-      /* istanbul ignore next */
       global.location.href = `${config.ECOMMERCE_BASE_URL}/payment/sdn/failure/`;
     }
-    throw new Error('SDN Failure');
+    throw new Error('This card holder did not pass the SDN check.');
```
robrap: What happens to this and this message?  Is it logged?  I assume not, but if it were, it should only be INFO.

src/payment/payment-methods/cybersource/service.js
```diff
@@ -19,36 +20,47 @@ export function configureApiService(newConfig, newApiClient) {
...
     `${config.ECOMMERCE_BASE_URL}/api/v2/sdn/search/`,
     {
       name: `${firstName} ${lastName}`,
       city,
       country,
     },
-  );
+  ).catch((error) => {
+    logAPIErrorResponse(error, {
+      messagePrefix: 'SDN Check Error',
```
robrap: As you did below, do we want to make it clear that this was an error attempting to check SDN, and not that the user didn't pass SDN?  This too can be in a follow-up PR.

src/payment/payment-methods/cybersource/service.js
abutterworth: This is not logged

PR #229, opened 2019-08-16
payment/data tests - 100% coverage!

davidjoy:


src/common/serviceUtils.js
```diff
@@ -12,34 +12,6 @@ export function applyConfiguration(expected, actual) {
...
- * }
- *
- * Into:
- *
- * {
- *   "name": "This value is invalid.",
- *   "other_field": "This other value is invalid"
- * }
- */
-export function unpackFieldErrors(fieldErrors) {
```
davidjoy: This was unused.

src/payment/cart/CouponForm.jsx
```diff
@@ -35,7 +35,7 @@ class CouponForm extends Component {
 
   handleRemoveSubmit(event) {
     event.preventDefault();
-    this.props.removeCoupon({ code: this.props.id });
+    this.props.removeCoupon({ id: this.props.id });
```
davidjoy: This was misleading - this is the voucher ID here, not the coupon code.  Updated it to remove ambiguity.

src/payment/data/__factories__/basket.factory.js
```diff
@@ -0,0 +1,46 @@
+import { Factory } from 'rosie'; // eslint-disable-line import/no-extraneous-dependencies
+import Chance from 'chance'; // eslint-disable-line import/no-extraneous-dependencies
+
+import './product.factory';
+import './message.factory';
+
+const chance = new Chance();
+
+Factory.define('basket')
```
davidjoy: This is a test factory!

src/payment/data/reducers.js
```diff
@@ -17,58 +17,55 @@ const basketInitialState = {
...
+      case updateQuantity.FULFILL:
+        return { ...state, isBasketProcessing: false };
+
+      default:
+    }
   }
+  return state;
 };
 
+/* istanbul ignore next */
```
davidjoy: Mocking universal-cookie isn't possible, far as I can tell. And node/jsdom doesn't have cookies.  So this code is basically untestable.

src/payment/data/sagas.js
```diff
@@ -10,34 +10,30 @@ import {
...
-  return yield select(state => state.payment.basket.isBasketProcessing);
-}
-
-export function* handleFetchBasket() {
+/**
+ * Many of the handlers here have identical error handling, and are also all processing the same
+ * sort of response object (a basket).  This helper is just here to dedupe that code, since its
+ * all identical.
+ */
+export function* catchBasketError(handler) {
```
davidjoy: I pulled this in from @MatthewPiatetsky 's PR (https://github.com/edx/frontend-app-payment/pull/67) and used it in all our handlers below.  Matthew, this may make your merge a bit interesting, but it also means it's already tested for you. :)

src/payment/data/sagas.js
```diff
@@ -47,64 +43,46 @@ export function* handleFetchBasket() {
   }
 }
 
+function* isBasketProcessing() {
+  return yield select(state => state.payment.basket.isBasketProcessing);
+}
+
+export function* handleFetchBasket() {
+  yield call(catchBasketError, function* processFetchBasket() {
```
davidjoy: These are much shorter now!

src/payment/data/selectors.js
```diff
@@ -47,7 +47,8 @@ export const paymentSelector = createSelector(
   queryParametersSelector,
   (basket, configuration, queryParameters) => {
     const isCouponRedeemRedirect =
-      queryParameters && queryParameters.coupon_redeem_redirect == 1; // eslint-disable-line eqeqeq
+      !!queryParameters &&
```
davidjoy: Turning the first part into a proper bool so it returns false if it's undefined/null.

src/payment/data/service.js
```diff
@@ -30,6 +30,7 @@ export function configureApiService(newConfig, newApiClient) {
 
   // For every ajax response, check if the API has
   // responded with a redirect value. If so, redirect.
+  /* istanbul ignore next */
```
davidjoy: I can't figure out how to test this.  So I punted.

src/payment/data/service.js
```diff
@@ -59,23 +60,10 @@ function getOrderType(productType) {
 }
 
 export function transformResults(data) {
-  const results = camelCaseObject(data) || {};
+  const results = camelCaseObject(data);
```
davidjoy: If data is `undefined` or `null`, I think it's better for us to blow up here than it is for us to blow up somewhere far away downstream when transformResults gives the app data that's sparse.

Ideally we might log something here and bail in a nice way, but this should now at least cause us to show a fallback-error.  Before it would just pass a really weird sparse "basket" back to the app.

src/payment/data/service.js
```diff
@@ -59,23 +60,10 @@ function getOrderType(productType) {
...
 
   const lastProduct = results.products && results.products[results.products.length - 1];
   results.orderType = getOrderType(lastProduct && lastProduct.productType);
-  results.orderTotal = Number(results.orderTotal);
```
davidjoy: So all of this isn't necessary; the data coming back are numbers already.  Line 73 below - regarding null providers - is deleted in @MatthewPiatetsky 's PR, so I proactively killed it here.  Once it's gone then the whole thing basically evaporates.

src/payment/data/sagas.js
abutterworth: Reads weird to me but definitely does the job.

PR #234, opened 2019-08-19
fix: add maxlength on form inputs

abutterworth:
Lengths are parity with legacy page. Should fix most problems with CR-1083.

Also changes autocomplete of country from country-name to just country (code).

### DIFF
--- a/src/payment/checkout/payment-form/CardHolderInformation.jsx
+++ b/src/payment/checkout/payment-form/CardHolderInformation.jsx
@@ -215,7 +215,7 @@ export class CardHolderInformationComponent extends React.Component {
               type="text"
               disabled={disabled}
               autocomplete="postal-code"
-              maxlength="10"
+              maxlength="9"
             />
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


src/payment/checkout/payment-form/CardHolderInformation.jsx
```diff
@@ -206,6 +215,7 @@ export class CardHolderInformationComponent extends React.Component {
               type="text"
               disabled={disabled}
               autocomplete="postal-code"
+              maxlength="10"
```
albemarle: Maybe constants?

src/payment/checkout/payment-form/CardHolderInformation.jsx
```diff
@@ -215,7 +215,7 @@ export class CardHolderInformationComponent extends React.Component {
               type="text"
               disabled={disabled}
               autocomplete="postal-code"
-              maxlength="10"
+              maxlength="9"
```
albemarle: What about the dash?

src/payment/checkout/payment-form/CardHolderInformation.jsx
abutterworth: We don't have any current map of fields/constants. I think moving these would be added organization or abstraction for little gain right now.

src/payment/checkout/payment-form/CardHolderInformation.jsx
abutterworth: This is matching the server constraints right now. If you put in 10 (even on the old page) you'll get an unhandled server error.

PR #235, opened 2019-08-19
fix: add basket processing specific action

abutterworth:
Rework processing into a single action so we can avoid reintroducing bugs regarding toggling the processing state moving forward

### DIFF
--- a/src/payment/data/sagas.js
+++ b/src/payment/data/sagas.js
@@ -3,6 +3,7 @@ import { call, put, takeEvery, select } from 'redux-saga/effects';
 // Actions
 import {
   basketDataReceived,
+  basketProcessing,
   fetchBasket,
   addCoupon,
   removeCoupon,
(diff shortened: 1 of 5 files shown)


src/payment/data/sagas.js
```diff
@@ -25,74 +26,56 @@ export const paymentMethods = {
...
+  return yield select(state => state.payment.basket.isBasketProcessing);
+}
+
 /**
  * Many of the handlers here have identical error handling, and are also all processing the same
  * sort of response object (a basket).  This helper is just here to dedupe that code, since its
  * all identical.
  */
-export function* catchBasketError(handler) {
+export function* performBasketOperation(operation, operationArgs) {
```
davidjoy: I think we should tweak this to:

```
export function* performBasketOperation(operation, ...operationArgs) {
```

src/payment/data/sagas.js
```diff
@@ -25,74 +26,56 @@ export const paymentMethods = {
...
+  if (yield isBasketProcessing()) return;
+
   try {
-    yield call(handler);
+    yield put(basketProcessing(true));
+    const result = yield call(operation, operationArgs);
```
davidjoy: And this to:

```
const result = yield call(operation, ...operationArgs);
```

PR #240, opened 2019-08-20
fix: bring card validator package into project

abutterworth:
Eliminate the dependency to shrink any risk of pulling in the third party library over time.


davidjoy:
Would it make more sense to fork their repository into the edx account and pull that in?  That'd keep the code out of this MFE.

nasthagiri:
We don't expect any other consumers of this code - so co-locating it inside this repo does make sense to me.  FYI @robrap

PR #249, opened 2019-08-23
Remove most of our mock stores.

davidjoy:
Replace them with real redux stores, actions, and test factories for source data.


src/feedback/__snapshots__/AlertList.test.jsx.snap
```diff
@@ -154,7 +154,7 @@ Array [
       <span>
         There was an unexpected problem. If the problem persists, please 
         <a
-          href="url"
+          href="http://localhost:18000/support"
```
davidjoy: Better source data.

src/payment/__factories__/basket.factory.js
```diff
@@ -4,7 +4,7 @@ import Chance from 'chance'; // eslint-disable-line import/no-extraneous-depende
 import './product.factory';
 import './message.factory';
 
-const chance = new Chance();
```
davidjoy: We're using the output of chance in snapshots now, so I'm seeding the RNG so it's always the same.

src/payment/__factories__/basket.factory.js
```diff
@@ -31,6 +30,7 @@ Factory.define('basket')
     }
     return products;
   })
+  .attr('summary_quantity', ['numProducts'], numProducts => numProducts)
```
davidjoy: This needed to match the number of products.

src/payment/__factories__/basket.factory.js
```diff
@@ -17,11 +17,10 @@ Factory.define('basket')
     order_total: 0.0,
     summary_subtotal: 0.0,
     summary_discounts: 0.0,
-    summary_quantity: 0,
     show_coupon_form: false,
     is_free_basket: false,
     currency: 'USD',
-    offers: null,
+    offers: [],
```
davidjoy: To my knowledge this was wrong.

src/payment/cart/order-details/data/selectors.js
```diff
@@ -22,7 +22,7 @@ const paymentSelector = (state, props) =>
 
 const lastProductSelector = createSelector(
   paymentSelector,
-  ({ products }) => (products ? products[products.length - 1] : undefined),
+  ({ products }) => products[products.length - 1],
```
davidjoy: The ternary operator here was handling a situation that never exists.  Removed it so I wouldn't have to write a test for the undefined clause.

src/payment/__factories__/basket.factory.js
abutterworth: Really? I thought quantity was the number of enrollment codes for a product?

PR #259, opened 2019-09-12
REV-940: transaction declined message error

julianajlk:
[REV-940](https://openedx.atlassian.net/browse/REV-940) 
- Updated MFE to display the message for TransactionDeclined (as opposed to receiving it from Ecommerce)

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "react-test-renderer": "16.9.0",
     "reactifex": "1.1.1",
     "redux-mock-store": "1.5.3",
-    "rosie": "2.0.1",
+    "rosie": "^2.0.1",
     "sass-loader": "7.2.0",
     "source-map-loader": "0.2.4",
     "style-loader": "0.23.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -7391,27 +7391,28 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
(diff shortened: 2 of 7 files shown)


src/payment/__snapshots__/PaymentPage.test.jsx.snap
```diff
@@ -10780,7 +10780,7 @@ exports[`<PaymentPage /> Renders correctly in various states should render the b
                 </span>
                 <span>
                   <span>
-                    $0.00
+                    -$0.00
```
abutterworth: Everything looks good except these two snapshots where $0.00 becomes -$0.00. Seems unrelated, but worth looking into.

abutterworth: Oh just checked travis. It looks like you just need to update these snapshots (maybe there was a problem before that is now fixed.)

julianajlk: Added `<LocalizedPrice amount={discounts ? discounts * -1 : 0} />` to prevent passing -0 to LocalizedPrice

package.json
```diff
@@ -115,7 +115,7 @@
     "react-test-renderer": "16.9.0",
     "reactifex": "1.1.1",
     "redux-mock-store": "1.5.3",
-    "rosie": "2.0.1",
+    "rosie": "^2.0.1",
```
robrap: @julianajlk: Why this change?  Thanks.

PR #269, opened 2019-10-03
[WIP] fix discount bug with payment error

MatthewPiatetsky:
issue: when a payment error happens and the basket page update the discount doesn't get reapplied. this causes the user to get charged the full amount, but then the receipt page shows they got charged the discounted amount

i don't think there is a way to test the fix locally so my suggestion for testing would be to 
merge, deploy to stage, modify local code so that an error on stage is possible, change flags on stage to direct everyone to mfe, make error purchase on stage? no sure if this will work  

if this doesn't work, then just retest successful purchases on stage, then deploy to prod, test payment errors and retest successful purchases and rollback if we find an issue

### DIFF
--- a/src/payment/data/sagas.js
+++ b/src/payment/data/sagas.js
@@ -163,6 +163,7 @@ export function* handleSubmitPayment({ payload }) {
       }
 
       if (error.basket) {
+        yield call(handleDiscountCheck);
         yield put(basketDataReceived(error.basket));
       }
     }
--- a/src/payment/data/sagas.test.js
+++ b/src/payment/data/sagas.test.js
@@ -739,10 +739,6 @@ describe('saga tests', () => {
         submitPayment.request(),
         clearMessages(),
         addMessage('ohboy', null, null, 'error'),
-        basketDataReceived({
-          i: 'am',
-          a: 'basket',
-        }),
         basketProcessing(false),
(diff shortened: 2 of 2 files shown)


src/payment/data/sagas.test.js
```diff
@@ -739,10 +739,6 @@ describe('saga tests', () => {
         submitPayment.request(),
         clearMessages(),
         addMessage('ohboy', null, null, 'error'),
-        basketDataReceived({
```
MatthewPiatetsky: don't know why these needed to be removed

src/payment/data/sagas.js
```diff
@@ -163,6 +163,7 @@ export function* handleSubmitPayment({ payload }) {
       }
 
       if (error.basket) {
+        yield call(handleDiscountCheck);
```
MatthewPiatetsky: if an error happens and the basket page updates the discount doesn't get reapplied

davidjoy: I don't think this will work quite right... handleDiscountCheck is going to dispatch the basketDataReceived action, and then on line 167 below we're going to blow away that data with the data in error.basket.

I think you want to do basketDataReceived on line 167 first, _then_ call handleDiscountCheck.  i.e., reverse the order of these two lines.

PR #271, opened 2019-10-04
fix discount bug with payment error

MatthewPiatetsky:


MatthewPiatetsky:
i'll need to go back after i manually verify that the issue is fixed to write a test

PR #272, opened 2019-10-07
Add purchased_for_organization basket attribute to bulk purchase form

Tj-Tracy:
Same change as before, now on the MFE:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7391,28 +7391,28 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
           "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -7391,28 +7391,28 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
```
crice100: Do we want to add all of these _resolved_ urls?

Tj-Tracy: This file is autogenerated by npm -- happens during minor version updates. It can be ignored.

src/payment/checkout/payment-form/CardHolderInformation.jsx
```diff
@@ -219,6 +219,26 @@ export class CardHolderInformationComponent extends React.Component {
...
+              <Field
+                id="purchasedForOrganization"
+                name="purchasedForOrganization"
+                component={FormInput}
+                type="checkbox"
+              />
+            </div>
+            <div className="col">
+              <label
+                htmlFor="purchasedForOrganization"
```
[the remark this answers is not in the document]
Tj-Tracy: No a11y issues that I've found, however I changed the label to contain a formatted message for i18n reasons.

PR #273, opened 2019-10-07
add some quick comments

MatthewPiatetsky:


davidjoy:
Not sure why the codecov/patch failed - it didn't change any actual code.

davidjoy:
(Put another way - I think we can ignore the codecov/patch failure and merge anyway)

PR #277, opened 2019-10-10
Add some speedcurve events to payment page

bbaker6225:
REV-984

Different types of events will be added later if this does not give us what we need.

Follows the basic idea of https://github.com/edx/edx-mktg/blob/4e3b18a6ad7e2a5b3a048d354f2fef56df0b474d/docroot/sites/all/themes/atedx/scripts/react/pages/privacy_policy/PrivacyPolicy.jsx#L11-#L18

### DIFF
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -129,3 +129,10 @@ export function generateAndSubmitForm(url, params = {}) {
   global.document.body.appendChild(form);
   form.submit();
 }
+
+// Used for SpeedCurve performance tracking
+export function markPerformanceIfAble(markName) {
+  if (performance.mark) {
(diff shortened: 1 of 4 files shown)


MatthewPiatetsky:
do you know if there's any way to test that this?  
or if the page renders without fail locally we could probably just merge and see if we see the events in speedcurve?
we can probably add some lines to make codecov ignore those lines or just merge and ignore the failure

bbaker6225:
I think the only way we can actually get anything useful out of this is to merge it, add the events to speedcurve, and see if they show up.

I can view the actual performance marks locally and they are _numbers_, but not useful ones by themselves

PR #280, opened 2019-10-15
Add segment events to accompany speedcurve events

bbaker6225:
Segment events will also likely be useful since we'll probably add
events to the backend as well.

### DIFF
--- a/src/payment/cart/OrderSummary.jsx
+++ b/src/payment/cart/OrderSummary.jsx
@@ -1,12 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { sendTrackEvent } from '@edx/frontend-analytics';
 import { FormattedMessage } from '@edx/frontend-i18n';
 
-import markPerformanceIfAble from '../speedcurve';
+import { markPerformanceIfAble, getPerformanceProperties } from '../performanceEventing';
 
 class OrderSummary extends React.Component {
   componentDidMount() {
     markPerformanceIfAble('Order Summary component rendered');
+    sendTrackEvent(
+      'edx.bi.ecommerce.payment_mfe.order_summary_rendered',
+      getPerformanceProperties(),
+    );
   }
 
   render() {
--- a/src/payment/performanceEventing.js
+++ b/src/payment/performanceEventing.js
@@ -1,0 +1,14 @@
+// Used for SpeedCurve performance tracking
+export function markPerformanceIfAble(markName) {
+  if (performance.mark) {
+    performance.mark(markName);
+  }
+}
+
+// Returns performance timing properties to send in events
+export function getPerformanceProperties() {
+  return {
+    emitted_at: Date.now() / 1000, // Seconds since epoch, for comparing timing on backend vs. frontend
+    performance_timestamp: performance.now(), // Milliseconds since time origin. See https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
+  };
+}
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,7 +9,7 @@ module.exports = createConfig('jest', {
     'src/index.jsx',
     'src/payment/checkout/card-validator',
     'src/payment/data/handleRequestError.js',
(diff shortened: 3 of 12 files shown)


bbaker6225:
Not in love with these event names, so let me know if you have better ideas

src/components/App.jsx
```diff
@@ -119,9 +120,10 @@ class App extends Component {
   }
 
   render() {
-    // Send performance event on initial render
+    // Send event on initial render
     if (!this.state.hasPreviouslyRendered) {
       utils.markPerformanceIfAble('Payment app began painting');
+      sendTrackEvent('edx.bi.ecommerce.payment_mfe.began_painting');
```
mulby: nit: maybe "started_painting"

src/payment/cart/OrderSummary.jsx
```diff
@@ -1,12 +1,14 @@
...
 import PropTypes from 'prop-types';
+import { sendTrackEvent } from '@edx/frontend-analytics';
 import { FormattedMessage } from '@edx/frontend-i18n';
 
 import { utils } from '../../common';
 
 class OrderSummary extends React.Component {
   componentDidMount() {
     utils.markPerformanceIfAble('Order Summary component rendered');
+    sendTrackEvent('edx.bi.ecommerce.payment_mfe.order_summary_rendered');
```
mulby: The timestamps on these events are pretty coarse, could we include the milliseconds from the start of the waterfall or something in the event?

bbaker6225: I'll add `performance.now()`. 

`The current high resolution time is the high resolution time from the time origin to the present time (typically called "now").`

I was mainly interested in the included timestamps for getting timing between the backend and frontend. Do you think they have high enough resolution for that?

mulby: I suspect not. Segment makes an HTTP request to their infrastructure when you call `analytics.track()`. The server records the timestamp of the event during the processing of that request. The clock from the client is not trusted.

For this reason, I suspect that two events fired very near one another (milliseconds apart) could actually be processed out-of-order.

bbaker6225: Hmph, ok thanks for the info. I suppose I can just do `Date.now() / 1000` and `time.time()` for the mfe and backend respectively, but that just feels _dirty_

src/common/utils.js
```diff
@@ -136,3 +136,11 @@ export function markPerformanceIfAble(markName) {
     performance.mark(markName);
   }
 }
+
+// Returns performance timing properties to send in analytics requests
+export function getPerformanceProperties() {
+  return {
+    emitted_at: Date.now() / 1000, // Seconds since epoch, for comparing timing on backend vs. frontend
```
bbaker6225: I'm not a fan of this but not sure of a better way to compare than using this and `time.time()`

src/components/App.jsx
```diff
@@ -119,9 +120,13 @@ class App extends Component {
   }
 
   render() {
-    // Send performance event on initial render
+    // Send event on initial render
     if (!this.state.hasPreviouslyRendered) {
       utils.markPerformanceIfAble('Payment app began painting');
+      sendTrackEvent(
```
bbaker6225: A concern I have here is that the `identify` call might not have been made yet, in which case we can't have any segment events here

src/payment/performanceEventing.js
```diff
@@ -0,0 +1,14 @@
...
+  if (performance.mark) {
+    performance.mark(markName);
+  }
+}
+
+// Returns performance timing properties to send in events
+export function getPerformanceProperties() {
+  return {
+    emitted_at: Date.now() / 1000, // Seconds since epoch, for comparing timing on backend vs. frontend
+    performance_timestamp: performance.now(), // Milliseconds since time origin. See https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
```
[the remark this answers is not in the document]
mulby: Got it.

PR #290, opened 2019-10-17
Remove PaymentForm loading performance mark

bbaker6225:
This has almost exactly the same timing as "Order Summary component
rendered", so let's get rid of this one and instead use "Payment Form
component rendered" to get a sense of when the right side of the page
comes in

In case we decide to close https://github.com/edx/frontend-app-payment/pull/280


bbaker6225:
Don't need this anymore

PR #291, opened 2019-10-17
Change Cart loading default to fix event

bbaker6225:
Make loading default to true so that OrderSummary doesn't render until
it actually is loaded.
Additionally has some eslint fixes that were not caught in travis


src/payment/performanceEventing.js
```diff
@@ -8,7 +8,11 @@ export function markPerformanceIfAble(markName) {
 // Returns performance timing properties to send in events
 export function getPerformanceProperties() {
   return {
-    emitted_at: Date.now() / 1000, // Seconds since epoch, for comparing timing on backend vs. frontend
-    performance_timestamp: performance.now(), // Milliseconds since time origin. See https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
+    // Seconds since epoch, for comparing timing on backend vs. frontend
+    emitted_at: Date.now() / 1000,
```
MatthewPiatetsky: any reason we're using seconds here vs milliseconds like below?

bbaker6225: I used seconds there to line up with `time.time()` on the backend, although I suppose I could make the backend line up with it here

PR #293, opened 2019-10-18
move optimizely snippet

MatthewPiatetsky:


package.json
```diff
@@ -55,7 +55,7 @@
     "lodash.pick": "4.4.0",
     "lodash.snakecase": "4.1.1",
     "lodash.without": "4.4.0",
-    "newrelic": "5.11.0",
+    "newrelic": "^5.13.1",
```
bbaker6225: Is this change needed for this?

MatthewPiatetsky: travis failed on me needing to run npm audit fix so i ran npm install and npm audit fix.  
don't really know if that's the correct process though or why the integrity stuff changed. think it might have to do with sha or something?

bbaker6225: Looks like the package `https-proxy-agent` needs to be manually upgraded (based off travis)

bbaker6225: https://github.com/codecov/codecov-node/issues/157 looks like we're waiting on codecov unfortunately

PR #302, opened 2019-10-28
Add Payment Page visual changes experiment

bbaker6225:
[REV-911](https://openedx.atlassian.net/browse/REV-911)

1.) Changes `Add Coupon Code (Optional)` into link that exposes coupon entry
2.) Removes the word `Required` from all fields, adds optional to 3 optional fields.  
3.) Removes unnecessary location fields
4.) The size of some fields is reduced
5.) Expiration Month Year changed to the `MM` `YY` format to match the back of a card.
The size of the card details fields are still pretty big, but it's tough to get them all on one line since our labels are all pretty long. Open to suggestions if people have ideas they think would look better

Ignore the right side of these screenshots. Updated right side below
**Coupon Input Closed**
Screenshot_2019-10-28 Payment edX(3)

**Coupon Input Open**
Screenshot_2019-10-28 Payment edX(1)

### DIFF
--- a/src/payment/checkout/payment-form/CardDetails.jsx
+++ b/src/payment/checkout/payment-form/CardDetails.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Field } from 'redux-form';
 import ReactTooltip from 'react-tooltip';
+import { Field } from 'redux-form';
 import { faLock, faQuestionCircle } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-i18n';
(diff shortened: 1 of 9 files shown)


src/payment/checkout/payment-form/CardHolderInformation.jsx
```diff
@@ -37,7 +37,180 @@ export class CardHolderInformationComponent extends React.Component {
   }
 
   render() {
-    const { disabled, showBulkEnrollmentFields } = this.props;
+    const { disabled, showBulkEnrollmentFields, isPaymentVisualExperiment } = this.props;
+
+    if (isPaymentVisualExperiment) {
```
bbaker6225: Originally thought I was going to be making more changes here that we decided not to do, so now it looks a bit overkill. This section is long, but basically just a temporary copy of the old section with different translations, required status, and some different ordering/formatting

bbaker6225:
Right side of page after removing state and postal code (since I got mixed up and thought they were required by the cybersource API)

Screenshot_2019-10-28 Payment edX(4)

julianajlk:
The credit card month/year have to be dropdowns?

src/payment/checkout/payment-form/CardDetails.jsx
```diff
@@ -54,8 +54,157 @@ export class CardDetailsComponent extends React.Component {
...
+      }
+      return items;
+    };
+
+    const yearText = this.props.intl.formatMessage(messages['payment.page.card.details.year']);
+    const currentYear = new Date().getFullYear();
+
+    return [
+      <option key="Year" value="">{yearText}</option>,
+      ...getYearOptions(currentYear, currentYear + 15),
```
MatthewPiatetsky: is +15 a standard amount?

bbaker6225: 1. Yes, tested both of those locally, but I think stage will be more useful for getting a more accurate idea of whether this really works
2. I'm lazy and use the `Squash and merge` button. The PR has the ticket link in the top level comment
3. Shelby's been out for a while, but maybe @emma-green can fill in. I believe the goal was to try and reach significance on revenue, but I agree that seems unlikely. If we want to try to add a bigger change, I could add card masking to this PR. Emma was thinking that card masking  might not need to be an experiment, but could just be added separately

bbaker6225: Oh and the `+15` is from https://github.com/edx/frontend-app-payment/blob/58a0aa018f4741118d2342757c6c2d0d4c1a05ac/src/payment/checkout/payment-form/CardDetails.jsx#L47-L55

Not sure why `+15`, but it seems pretty reasonable to me

MatthewPiatetsky: 1. can't you just test against stage using https://local.stage.edx.org:1998?
3. i think the card mask goes into the category of good UX features that probably won't shift revenue, so i doubt that would address the issue of the sum total changes being potentially too small. Not necessarily a blocker to merging, but something we should have an answer to before starting the experiment.

bbaker6225: 1. The local run stage stuff has never worked for me unfortunately
2. I'll let Emma and product decide. If this stuff doesn't give a significant impact though then maybe we either shouldn't do it or just add it anyway if we're really attached

MatthewPiatetsky: 1. ah interesting, i've only ever used that locally actually, but sure i guess testing on stage could work then
2. yup their decision, though my take on it would probably be that we could either add even more changes/add back some of the scrapped changes, consider just implementing these changes without a test or consider implementing them with a test even if it doesn't reach significance for 5% or over

PR #303, opened 2019-10-29
Fix issue with state required for US

bbaker6225:
Fixes issue where state/province was still required for US/Canada

### DIFF
--- a/src/payment/checkout/payment-form/CardHolderInformation.jsx
+++ b/src/payment/checkout/payment-form/CardHolderInformation.jsx
@@ -145,7 +145,6 @@ export class CardHolderInformationComponent extends React.Component {
                 component={FormSelect}
                 options={this.renderCountryOptions()}
                 required
-                onChange={this.handleSelectCountry}
                 disabled={disabled}
                 autoComplete="country"
               />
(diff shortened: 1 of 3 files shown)


src/payment/checkout/payment-form/CardHolderInformation.jsx
```diff
@@ -145,7 +145,6 @@ export class CardHolderInformationComponent extends React.Component {
                 component={FormSelect}
                 options={this.renderCountryOptions()}
                 required
-                onChange={this.handleSelectCountry}
```
bbaker6225: https://github.com/edx/frontend-app-payment/blob/11e13aa332eb974ac397f20a62add9bfcc325e15/src/payment/checkout/payment-form/CardHolderInformation.jsx#L21-L24 is only used by https://github.com/edx/frontend-app-payment/blob/11e13aa332eb974ac397f20a62add9bfcc325e15/src/payment/checkout/payment-form/CardHolderInformation.jsx#L336-L337
which is not a field in the experiment, so I'm removing the handler for the experiment just to make sure that's not connected to what I'm seeing

src/payment/checkout/payment-form/PaymentForm.jsx
```diff
@@ -128,7 +128,7 @@ export class PaymentFormComponent extends React.Component {
       };
     }
 
-    if (getStates(country)) {
+    if (getStates(country) && !this.props.isPaymentVisualExperiment) {
```
bbaker6225: Accidentally discarded this change from my old PR

PR #305, opened 2019-10-29
Update form to be editable on load

julianajlk:
[REV-937](https://openedx.atlassian.net/browse/REV-937).
Enable form fields to be editable while basket is loading.
Disable form fields onSubmit of form. 
Screen Shot 2019-10-29 at 3 41 09 PM
Screen Shot 2019-10-29 at 3 49 40 PM

Bulk Purchase - quantity updating disable submit button. 
Screen Shot 2019-10-31 at 10 34 27 AM
*Note: updated CSS of checkbox, on zoom in Chrome the size changes to an abnormally large size.

### DIFF
--- a/src/payment/checkout/Checkout.jsx
+++ b/src/payment/checkout/Checkout.jsx
@@ -73,8 +73,9 @@ class Checkout extends React.Component {
     const {
       intl,
       isFreeBasket,
-      loading,
       isBasketProcessing,
+      loading,
+      loaded,
       isPaymentVisualExperiment,
       paymentMethod,
       submitting,
@@ -83,6 +84,7 @@ class Checkout extends React.Component {
 
     const submissionDisabled = loading || isBasketProcessing;
     const isBulkOrder = orderType === ORDER_TYPES.BULK_ENROLLMENT;
+    const isQuantityUpdating = isBasketProcessing && loaded;
 
     // istanbul ignore next
     const payPalIsSubmitting = submitting && paymentMethod === 'paypal';
@@ -130,10 +132,11 @@ class Checkout extends React.Component {
         <PaymentForm
           onSubmitPayment={this.handleSubmitCybersource}
           onSubmitButtonClick={this.handleSubmitCybersourceButtonClick}
-          disabled={submissionDisabled}
+          disabled={submitting}
           loading={loading}
           isProcessing={cybersourceIsSubmitting}
           isBulkOrder={isBulkOrder}
+          isQuantityUpdating={isQuantityUpdating}
           isPaymentVisualExperiment={isPaymentVisualExperiment}
         />
       </React.Fragment>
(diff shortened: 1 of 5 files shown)


src/payment/checkout/Checkout.jsx
```diff
@@ -74,14 +74,13 @@ class Checkout extends React.Component {
...
       loading,
-      isBasketProcessing,
       isPaymentVisualExperiment,
       paymentMethod,
       submitting,
       orderType,
     } = this.props;
 
-    const submissionDisabled = loading || isBasketProcessing;
+    const submissionDisabled = submitting;
```
abutterworth: I think achieving the editability of form fields this way technically would allow form submission during load. This is unlikely to be an issue,  but an alternative may be to create a new `editable` prop instead of using `submissionDisabled` as the way to disable form fields.

julianajlk: Is form submission possible even when the submit button is disabled?

bbaker6225: Only through someone editing the HTML or something like that. It's always going to be possible to submit the form client side, but the backend handles that so I don't think it's a big concern

julianajlk: I will disable the submit button on bulk purchases to prevent the above issue @bbaker6225, thank you!

src/payment/checkout/Checkout.jsx
```diff
@@ -73,16 +73,19 @@ class Checkout extends React.Component {
...
+      loading,
+      loaded,
       isPaymentVisualExperiment,
       paymentMethod,
       submitting,
       orderType,
     } = this.props;
 
     const submissionDisabled = loading || isBasketProcessing;
+    const editable = submitting;
```
bbaker6225: It's counter-intuitive to me that we would want to disable the form when it is `editable` by its current definition. I think we probably want to come up with a new definition of editable here, or just not use it

src/payment/checkout/Checkout.jsx
```diff
@@ -130,10 +133,11 @@ class Checkout extends React.Component {
         <PaymentForm
           onSubmitPayment={this.handleSubmitCybersource}
           onSubmitButtonClick={this.handleSubmitCybersourceButtonClick}
-          disabled={submissionDisabled}
+          disabled={editable}
           loading={loading}
           isProcessing={cybersourceIsSubmitting}
           isBulkOrder={isBulkOrder}
+          isQuantityUpdating={bulkQuantityIsUpdating}
```
bbaker6225: Nit: Personally, if we choose to use this prop, I would name it the same thing in both `PaymentForm` and `Checkout`

julianajlk: Noted, will do. I followed the `cybersourceIsSubmitting` but also prefer naming the same.

src/payment/checkout/Checkout.jsx
julianajlk: Thanks for seeing that, at some point I had more than one condition, added `submitting` as props without having the separate variable!

PR #309, opened 2019-10-31
Add card masking to visual experiment

bbaker6225:
[REV-911](https://openedx.atlassian.net/browse/REV-911)

Supported cards come from: https://support.edx.org/hc/en-us/articles/206502498-How-do-I-pay-
Card formats taken from http://wysiwyg.co.il/Anatomy-of-CreditCard-Number-formats.asp and just searching around

**Visa Format**
Screenshot_2019-10-31 Payment edX

**AMEX Format**:
Screenshot_2019-10-31 Payment edX(1)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7894,8 +7894,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -7916,14 +7915,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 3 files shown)


src/payment/checkout/payment-form/CardDetails.jsx
```diff
@@ -22,6 +23,24 @@ export class CardDetailsComponent extends React.Component {
     };
   }
 
+  componentDidMount() {
+    if (this.props.isPaymentVisualExperiment) {
+      const cardFormats = [
+        '4999 9{4} 9{4} 9{4}', // Visa
```
bbaker6225: For `inputmask`, `9` represents any number and `{4}` represents 4 of the same thing. So this line corresponds to:
`4xxx xxxx xxxx xxxx` where `x` is any number

bbaker6225:
@MatthewPiatetsky Interesting point. That's probably more a question for legal, since I believe similar PCI concerns were what led to the card validator package being moved inside this repo

package-lock.json
```diff
@@ -7916,14 +7915,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
```
julianajlk: Not sure what update this refers to but should this be in the commit?

bbaker6225:
I have reached out to legal to see if there are any PCI concerns with this

package-lock.json
bbaker6225: I know it's an unsatisfying answer, but this is just what I get from running `npm install inputmask`. 🤷‍♂

bbaker6225:
We are going to hold off on adding this to the experiment for now, pending an in-depth PCI review of the package

PR #315, opened 2019-11-04
build: fix stage-dev server

abutterworth:
Upgrade frontend-build which now includes a dev-stage webpack config to build from. `npm run start:stage` should now work properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "universal-cookie": "4.0.2"
   },
   "devDependencies": {
-    "@edx/frontend-build": "^1.3.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -75,7 +75,7 @@
     "universal-cookie": "4.0.2"
   },
   "devDependencies": {
-    "@edx/frontend-build": "1.2.2",
+    "@edx/frontend-build": "^1.3.1",
```
bbaker6225: Oh one comment. Do we want to pin this to be in line with everything else?

abutterworth: All set.

PR #324, opened 2019-11-20
Add num enrolled experiment

julianajlk:
[REV-1013](https://openedx.atlassian.net/browse/REV-1013).

Adding Optimizely Experiment to display:
- Number enrolled of a course. Excludes courses with < 5000 enrolled. 

Screen Shot 2019-11-22 at 2 38 38 PM

Note: enrollmentCountData content will be added in Optimizely.

### DIFF
--- a/src/payment/PaymentPage.jsx
+++ b/src/payment/PaymentPage.jsx
@@ -27,11 +27,15 @@ class PaymentPage extends React.Component {
     const {
       experimentVariables: {
         isPaymentVisualExperiment = false,
+        isNumEnrolledExperiment = false,
+        enrollmentCountData = [],
       } = {},
     } = window;
 
     this.state = {
       isPaymentVisualExperiment,
+      isNumEnrolledExperiment,
+      enrollmentCountData,
     };
   }
 
@@ -42,7 +46,11 @@ class PaymentPage extends React.Component {
 
   renderContent() {
     const { isEmpty, isRedirect } = this.props;
-    const { isPaymentVisualExperiment } = this.state;
+    const {
+      isPaymentVisualExperiment,
+      isNumEnrolledExperiment,
+      enrollmentCountData,
+    } = this.state;
 
     // If we're going to be redirecting to another page instead of showing the user the interface,
     // show a minimal spinner while the redirect is happening.  In other cases we want to show the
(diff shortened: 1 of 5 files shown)


src/payment/cart/ProductLineItem.jsx
```diff
@@ -4,6 +4,17 @@ import { FormattedMessage, injectIntl } from '@edx/frontend-i18n';
 
 
 class ProductLineItem extends React.PureComponent {
+  renderEnrollmentCount(courseKey, enrollmentCountData) {
+    const courseObj = enrollmentCountData.find(course => course.key === courseKey);
+    const enrollmentCount = courseObj.enrollment_count;
+    if (courseObj !== undefined) {
+      return (
+        <p className="num-enrolled"><strong>{enrollmentCount}</strong> already enrolled!</p>
```
bbaker6225: Do we want to translate this?

src/payment/PaymentPage.jsx
```diff
@@ -27,11 +27,15 @@ class PaymentPage extends React.Component {
     const {
       experimentVariables: {
         isPaymentVisualExperiment = false,
+        isNumEnrolledExperiment = false,
```
bbaker6225: I would personally use just or the other of these experiment variables, assuming all the changes are one experiment. Not a big deal either way

src/payment/PaymentPage.jsx
```diff
@@ -27,11 +27,15 @@ class PaymentPage extends React.Component {
...
+        enrollmentCountData = [],
```
bbaker6225: Are you adding this data via optimizely? I guess that's fine if Shelby said we don't want to put the full json in the code

src/payment/cart/ProductLineItem.jsx
```diff
@@ -4,6 +4,17 @@ import { FormattedMessage, injectIntl } from '@edx/frontend-i18n';
 
 
 class ProductLineItem extends React.PureComponent {
+  renderEnrollmentCount(courseKey, enrollmentCountData) {
+    const courseObj = enrollmentCountData.find(course => course.key === courseKey);
+    const enrollmentCount = courseObj.enrollment_count;
```
bbaker6225: I think this should go in the if statement

src/payment/PaymentPage.jsx
julianajlk: Thought it would be better to keep it separate in case there is another experiment that requires to turn on only one of them

src/payment/PaymentPage.jsx
julianajlk: I am adding via Optimizely to the window Object. Matthew suggested to add this as a JSON file in PaymentMFE since I am making changes might as well add it to the code

PR #326, opened 2019-11-20
Force refresh the jwt cookie before all post requests

abutterworth:
Related to ARCH-1304.

Some users payment submissions fail with a jwt expiration error on the server. It's unclear why the this client application doesn't see the jwt as expire too. We will investigate. Temporarily this fix will refresh the jwt cookie before every post request on the page to ensure the jwt cookie is as fresh as it could be. We will see if this eliminates the errors.

Also, this includes upgrades to satisfy npm audit.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8831,9 +8831,9 @@
       "dev": true
     },
     "handlebars": {
-      "version": "4.5.1",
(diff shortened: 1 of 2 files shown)


abutterworth:
> 2\. Do you know that this interceptor is run first, or does it not really matter anyway?

It doesn't matter, but I think this is run first (since it's added after frontend-auth's)

PR #329, opened 2019-11-22
feat: upgrade to frontend-platform

abutterworth:
Replace all frontend runways with the new, consolidated frontend-platform. This includes a variety of API changes, the core of which is that `frontend-` `auth`, `i18n`, `logging`, `analytics` functionality is now imported from `frontend-platform/{pkg}`


package.json
```diff
@@ -35,13 +35,9 @@
...
-    "@edx/frontend-analytics": "3.0.0",
-    "@edx/frontend-auth": "7.0.1",
-    "@edx/frontend-base": "4.1.0",
-    "@edx/frontend-component-footer": "9.0.0",
-    "@edx/frontend-component-header": "1.1.4",
-    "@edx/frontend-i18n": "3.0.3",
-    "@edx/frontend-logging": "3.0.1",
+    "@edx/frontend-component-footer": "^10.0.0",
+    "@edx/frontend-component-header": "^2.0.0",
+    "@edx/frontend-platform": "^1.1.1",
```
davidjoy: 1.1.3

abutterworth: upgraded.

src/index.jsx
```diff
@@ -17,7 +28,29 @@ import configureStore from './data/configureStore';
...
+  APPLE_PAY_MERCHANT_NAME: process.env.APPLE_PAY_MERCHANT_NAME,
+  APPLE_PAY_COUNTRY_CODE: process.env.APPLE_PAY_COUNTRY_CODE,
+  APPLE_PAY_CURRENCY_CODE: process.env.APPLE_PAY_CURRENCY_CODE,
+  APPLE_PAY_START_SESSION_URL: process.env.APPLE_PAY_START_SESSION_URL,
+  APPLE_PAY_AUTHORIZE_URL: process.env.APPLE_PAY_AUTHORIZE_URL,
+  APPLE_PAY_SUPPORTED_NETWORKS: process.env.APPLE_PAY_SUPPORTED_NETWORKS && process.env.APPLE_PAY_SUPPORTED_NETWORKS.split(','),
+  APPLE_PAY_MERCHANT_CAPABILITIES: process.env.APPLE_PAY_MERCHANT_CAPABILITIES && process.env.APPLE_PAY_MERCHANT_CAPABILITIES.split(','),
+});
+
+subscribe(APP_ANALYTICS_INITIALIZED, () => {
```
davidjoy: We're  putting this code into an analytics-phase handler in frontend-platform, so this whole subscribe block should be deletable.

src/payment/AlertCodeMessages.jsx
```diff
@@ -1,9 +1,14 @@
...
 import PropTypes from 'prop-types';
-import { FormattedMessage } from '@edx/frontend-i18n';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import { Hyperlink } from '@edx/paragon';
 import LocalizedPrice from './cart/LocalizedPrice';
-import { App } from '@edx/frontend-base';
+import { getConfig, mergeConfig } from '@edx/frontend-platform';
+
+
+mergeConfig({
```
davidjoy: This is duplicative of the `mergeConfig`in index.jsx above.   I think this one can be deleted.

src/payment/AlertCodeMessages.test.jsx
```diff
@@ -15,7 +15,9 @@ const mockStore = configureMockStore();
 describe('SingleEnrollmentCodeWarning', () => {
   it('should render with values', () => {
     const component = (
-      <SingleEnrollmentCodeWarning values={{ courseAboutUrl: 'http://edx.org' }} />
+      <IntlProvider locale="en">
```
davidjoy: Surprised adding IntlProvider here was necessary... did it improve the snapshot to do so?

src/payment/PaymentPage.test.jsx
```diff
@@ -35,36 +41,61 @@ jest.mock('universal-cookie', () => {
...
-    const userAccount = Factory.build('userAccount');
-
     store = createStore(createRootReducer(), {}, applyMiddleware(thunkMiddleware));
-    store.dispatch(fetchUserAccountSuccess(userAccount));
   });
 
   describe('Renders correctly in various states', () => {
     beforeEach(() => {
       analytics.sendTrackingLogEvent = jest.fn();
-      Cookies.result = undefined;
```
davidjoy: Deleting this seems wrong - it means the cookie will exist for these tests, suddenly, when their default state was apparently supposed to be one where they had no currency/language cookies (undefined)

src/payment/AlertCodeMessages.jsx
abutterworth: It's added so tests work. I could put this code in the test itself. wdyt?

src/payment/PaymentPage.test.jsx
```diff
@@ -88,22 +121,19 @@ describe('<PaymentPage />', () => {
       expect(tree.toJSON()).toMatchSnapshot();
     });
 
-    it('should render the basket in a different currency', () => {
-      Cookies.result = {
```
davidjoy: Given the new default above, deleting this `Cookies.result =` feels right.

src/payment/PaymentPage.test.jsx
```diff
@@ -117,9 +147,11 @@ describe('<PaymentPage />', () => {
     it('should render the basket with an enterprise offer', () => {
       const component = (
         <IntlProvider locale="en">
-          <Provider store={store}>
-            <PaymentPage />
-          </Provider>
+          <AppContext.Provider value={{ authenticatedUser }}>
```
davidjoy: Is it the case that all these usages of Provider don't rely on the `config` too?  We're only giving AppContext part of its normal values here... (locale is missing too)

src/payment/AlertCodeMessages.test.jsx
abutterworth: Adding it eliminated warnings in the console when running the test.
```
console.error node_modules/react-intl/lib/index.js:1533
    [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry. Using default message as fallback.
```

src/payment/__snapshots__/PaymentPage.test.jsx.snap
```diff
@@ -8727,10 +8727,7 @@ exports[`<PaymentPage /> Renders correctly in various states should render the b
               </span>
               <span>
                 <span>
-                  <span>
```
davidjoy: So that this changed is concerning - doesn't it mean that it used to have a currency cookie for MX and now it doesn't?

davidjoy: This snapshot says "should render the basket in a different currency"

src/payment/__snapshots__/PaymentPage.test.jsx.snap
```diff
@@ -8747,23 +8744,9 @@ exports[`<PaymentPage /> Renders correctly in various states should render the b
               <span
                 className="text-right"
               >
-                <span>
```
davidjoy: Ditto here.

src/payment/__snapshots__/PaymentPage.test.jsx.snap
abutterworth: Good catch. looking...

src/payment/cart/order-details/data/selectors.js
```diff
@@ -16,13 +16,6 @@ import {
 
 export const storePath = ['payment', 'basket'];
 
-export const userAccountSelector = state => state.userAccount;
```
davidjoy: Awwww yiiiiis

src/payment/checkout/Checkout.test.jsx
```diff
@@ -11,11 +11,37 @@ import '../__factories__/basket.factory';
...
+    he: {},
+    ca: {},
+    'zh-cn': {},
+    fr: {},
+    'es-419': {},
+    ar: {}
+  },
+});
+
+import { sendTrackEvent } from '@edx/frontend-platform/analytics'; // eslint-disable-line import/first
```
davidjoy: I think in a recent version of jest they added handling to hoist `jest.mock` calls to the top of the scope, which made it unnecessary to import things "after" they're `jest.mock`'d.  Point being, we could move this up and remove the eslint-disable-line comment if we wanted.

src/payment/PaymentPage.test.jsx
abutterworth: Added back in in a form – the cookie mock mechanism was updated some.

src/payment/data/service.js
```diff
@@ -78,17 +79,17 @@ export async function postCoupon(code) {
...
 }
 
 export async function getDiscountData(courseKey) {
-  const { data } = await App.apiClient.get(
-    `${App.config.LMS_BASE_URL}/api/discounts/course/${courseKey}`,
+  const { data } = await getAuthenticatedHttpClient().get(
+    `${getConfig().LMS_BASE_URL}/api/discounts/course/${courseKey}`,
     {
-      xhrFields: { withCredentials: true },
+      withCredentials: true,
```
davidjoy: Sanity check - this `xhrFields` thing was unnecessary?  Or was it wrong?

src/payment/PaymentPage.test.jsx
abutterworth: Added back in.

src/payment/payment-methods/apple-pay/service.js
```diff
@@ -1,7 +1,19 @@
-import { App } from '@edx/frontend-base';
-import { logApiClientError } from '@edx/frontend-logging';
+import { ensureConfig, getConfig, mergeConfig } from '@edx/frontend-platform';
+import { logError } from '@edx/frontend-platform/logging';
+import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
-App.ensureConfig([
+
+mergeConfig({
+  APPLE_PAY_MERCHANT_NAME: process.env.APPLE_PAY_MERCHANT_NAME,
```
davidjoy: Isn't all of this merging done in index.jsx?

davidjoy: (it only needs to be done once)

src/payment/payment-methods/apple-pay/service.js
```diff
@@ -107,7 +119,7 @@ export function checkout(basket) {
         .catch((error) => {
           // eslint-disable-next-line no-param-reassign
           error.code = 'apple-pay-authorization-failure';
-          logApiClientError(error, {
+          logError(error, {
```
davidjoy: Does logError accept a messagePrefix still?  Looks like it doesn't do anything special with it, it'll just pass it through like any other `customAttributes` key.  Is that okay?

src/payment/payment-methods/cybersource/service.js
```diff
@@ -1,11 +1,16 @@
...
 import formurlencoded from 'form-urlencoded';
-import { logApiClientError } from '@edx/frontend-logging';
+import { logError } from '@edx/frontend-platform/logging';
 
 import handleRequestError from '../../data/handleRequestError';
 import { generateAndSubmitForm } from '../../data/utils';
 
-App.ensureConfig([
+mergeConfig({
+  CYBERSOURCE_URL: process.env.CYBERSOURCE_URL,
```
davidjoy: Ditto

src/payment/payment-methods/cybersource/service.js
```diff
@@ -18,15 +23,15 @@ App.ensureConfig([
...
+    `${getConfig().ECOMMERCE_BASE_URL}/api/v2/sdn/search/`,
     {
       name: `${firstName} ${lastName}`,
       city,
       country,
     },
   ).catch((error) => {
-    logApiClientError(error, {
+    logError(error, {
       messagePrefix: 'SDN Check Error',
```
davidjoy: Another messagePrefix - partially I'm just worried that the fact we're not actually using it as a prefix will break some reporting on the other side

src/payment/__snapshots__/PaymentPage.test.jsx.snap
abutterworth: Fixed

src/payment/payment-methods/cybersource/service.js
davidjoy: Going to stop commenting on individual messagePrefix instances

src/payment/AlertCodeMessages.jsx
davidjoy: Ooohhh, I see.  Huh.  I wonder if you could do it once in setupTest to mirror what happens in index.jsx?

src/payment/AlertCodeMessages.test.jsx
davidjoy: 👍

src/payment/checkout/Checkout.test.jsx
abutterworth: updated.

src/payment/data/service.js
abutterworth: It's jQuery syntax.

davidjoy:
Pre-emptive approval, pending sanity checks and nits and such.

src/payment/payment-methods/apple-pay/service.js
abutterworth: yes, this like above is here to make tests function. I could add these lines into the test instead.

src/payment/payment-methods/apple-pay/service.js
abutterworth: It doesn't do anything special with it. We can make a subsequent update to frontend-platform to prepend messagePrefixes if we want. It's one of the improvements Robert had suggested for logging I think

src/payment/payment-methods/cybersource/service.js
abutterworth: This was stuff I'd added. I don't there's any reporting hooked up to search for these error messages. AFAIK

src/index.jsx
abutterworth: Swapped

src/payment/AlertCodeMessages.jsx
abutterworth: Oh wow yea. I'll do that.

src/payment/payment-methods/apple-pay/service.js
abutterworth: config is now merged in setupTest

src/payment/payment-methods/cybersource/service.js
abutterworth: config is now merged in setupTest

src/payment/PaymentPage.test.jsx
abutterworth: Looks like not, but I added locale and config for completeness.

PR #330, opened 2019-11-26
Removed i18n from num enrolled

julianajlk:
[REV-1013](https://openedx.atlassian.net/browse/REV-1013). 
Removing i18n because the test is only for users with English language settings.


julianajlk:
Accidentally closed this PR.

PR #345, opened 2019-12-05
[REV-1045] Decrease perception of loading time on left side

MatthewPiatetsky:
before:
loading2
after:
loading3

### DIFF
--- a/src/payment/cart/Cart.jsx
+++ b/src/payment/cart/Cart.jsx
@@ -28,6 +28,7 @@ class Cart extends React.Component {
       isCurrencyConverted,
       isPaymentVisualExperiment,
       isNumEnrolledExperiment,
+      REV1045Experiment,
       enrollmentCountData,
       orderTotal,
       showCouponForm,
@@ -36,6 +37,7 @@ class Cart extends React.Component {
       summaryQuantity,
       summaryDiscounts,
       offers,
+      loaded
     } = this.props;
 
     const isBulkOrder = orderType === ORDER_TYPES.BULK_ENROLLMENT;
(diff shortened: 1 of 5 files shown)


emma-green:
Perhaps I missed something, but what's the point of not showing the skeleton? Wouldn't that increase the perception of loading time?

MatthewPiatetsky:
@emma-green well right now the process is
1. page loads with skeleton on left side
2. wait for full left side to load
3. show page

I changed it to
1. Page loads with in your cart
2. Product details appears when it is ready
3. Order Details appears when it is ready
4. Order summary appears when it is ready

The skeleton is rather complex, so adding a skeleton for each of those 3 sections would take several days. For juliana's suggestion, we could potentially only add a single spinner on the order summary, that would certainly be simpler. We could certainly add the skeleton for each section, would just take a while.
Could you clarify your point though about how this would increase perception of loading time?
Things popping in when they're ready appears to load  quicker i think?

src/payment/cart/Cart.jsx
```diff
@@ -78,9 +89,11 @@ class Cart extends React.Component {
             <CouponForm isPaymentVisualExperiment={isPaymentVisualExperiment} /> : null}
           <TotalTable total={orderTotal} />
           {isCurrencyConverted ? <CurrencyDisclaimer /> : null}
-        </OrderSummary>
+        </OrderSummary> : (<div><div className="skeleton py-2 mb-3 w-50" />
```
julianajlk: You can also add <React.Fragment></React.Fragment> if you don't want to add the extra < div> here, so you can still group multiple children without adding the extra node.

src/payment/cart/order-details/OrderDetails.jsx
```diff
@@ -85,7 +85,10 @@ class OrderDetails extends Component {
 
   render() {
     if (this.props.messageType === null) {
-      return null;
+      return (this.props.REV1045Experiment ? <div><div className="skeleton py-2 mb-3 w-50" />
```
julianajlk: Same as the above, could replace < div> with <React.Fragment>

PR #347, opened 2019-12-10
build: turn on auto-merge for Renovate patches.

davidjoy:


davidjoy:
Because automerge is nested inside of “patch”, it’ll only apply to patch-level updates. M.m.p - Major and minor will still require manual review. The idea is that bug fixes will be merged automatically, while features and breaking changes should be reviewed. 

One possible concern for the payment MFE is whether we really want this to happen automatically. It’s possible that for PCI reasons, we want to manually review all updates and not rely on a bot. @emma-green, do I have that right?  Feel free to close or merge depending on the right call.

emma-green:
I will discuss with the team today.

PR #355, opened 2020-01-29
Add transparent pricing experiment

julianajlk:
[REV-1062](https://openedx.atlassian.net/browse/REV-1062).
- Experiment to add a fair price messaging to Payment MFE. 

Screen Shot 2020-01-29 at 1 58 17 PM
Screen Shot 2020-01-29 at 1 58 10 PM

Screen Shot 2020-01-29 at 4 23 37 PM Screen Shot 2020-01-29 at 4 23 44 PM

### DIFF
--- a/src/payment/PaymentPage.jsx
+++ b/src/payment/PaymentPage.jsx
@@ -30,6 +30,7 @@ class PaymentPage extends React.Component {
         isNumEnrolledExperiment = false,
         REV1045Experiment = false,
         isPriceMessageExperiment = false,
+        isTransparentPricingExperiment = false,
         enrollmentCountData = [],
       } = {},
     } = window;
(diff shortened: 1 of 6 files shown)


src/payment/PaymentPage.jsx
```diff
@@ -30,6 +30,7 @@ class PaymentPage extends React.Component {
         isNumEnrolledExperiment = false,
         REV1045Experiment = false,
         isPriceMessageExperiment = false,
+        isTransparentPricingExperiment = false,
```
MatthewPiatetsky: not sure how much work it would be or if it would be feasible, but imo it would be better if we stopped hardcoding the variables for each experiment and just checked the experimentVariables dictionary directly

julianajlk: @MatthewPiatetsky I agree, or even do an experiment clean up and remove the experiments that did not reach significance/not planned to be used again. To answer your other questions - 
1. I've added the mobile view on the description. 
2. We're not translating for this test, it will be for all learners. 
3. Came from product, not sure how they got to that number.

MatthewPiatetsky: would it still work if you didn't hardcode the isTransparentPricingExperiment variable and just passed through the experimentVariables dict?

PR #356, opened 2020-02-18
[REV-1074] Copy static checkout pages 

MatthewPiatetsky:
to the dist folder where they will be accessible through a url

This is necessary because the html for those pages needs to be in the dist folder to be publicly accessible
https://openedx.atlassian.net/browse/REV-1112
@edx/revenue-squad

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "axios-mock-adapter": "1.17.0",
     "chance": "1.1.0",
     "clean-webpack-plugin": "3.0.0",
+    "copy-webpack-plugin": "^5.1.1",
     "dotenv-webpack": "1.7.0",
     "enzyme": "3.10.0",
     "enzyme-adapter-react-16": "1.14.0",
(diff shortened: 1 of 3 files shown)


webpack.dev-stage.config.js
```diff
@@ -13,6 +14,13 @@ const config = getBaseConfig('webpack-dev-stage');
  * - We need to modify the path for Dotenv to use our .env.development-stage file.
  */
 config.plugins = [
+  // Copies over static html checkout pages to dist folder for the REV-1074 experiment
+  new CopyPlugin([
```
abutterworth: This is unneeded I believe.

webpack.dev.config.js
```diff
@@ -12,6 +13,13 @@ const config = getBaseConfig('webpack-dev');
  * - We want a custom html-webpack-plugin config
  */
 config.plugins = [
+  // Copies over static html checkout pages to dist folder for the REV-1074 experiment
+  new CopyPlugin([
```
abutterworth: This is unneeded I believe.

MatthewPiatetsky:
@julianajlk that was a sku locally so I could test the redirection. once we generate the files we'll want them to be named <sku>.html

PR #358, opened 2020-02-21
Add Static Checkout page 

julianajlk:
[REV-1111](https://openedx.atlassian.net/browse/REV-1111). 
WIP on the HTML/CSS and JS files for the static checkout page.


public/static-checkout/css/style.css
```diff
@@ -0,0 +1,256 @@
...
+  display: table ;
+  clear:both;
+}
+
+[class*='col-'] {
+  display: inline-block;
+  min-height: 1px;
+}
+
+.col-1 {width: 8.33%;}
```
MatthewPiatetsky: it looks like for now most of these classes are unused and could be removed? (unless they become necessary for making things responsive)

public/static-checkout/index.html
```diff
@@ -0,0 +1,148 @@
+<!DOCTYPE html>
+
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Payment | edX</title>
+  <link rel="shortcut icon" type="image/x-icon" href="assets/favicon.ico"/>
+  <link href="css/style.css" rel="stylesheet" type="text/css">
```
MatthewPiatetsky: this should probably be minified and directly included here, but we could do that when we're ready

public/static-checkout/css/style.css
```diff
@@ -0,0 +1,256 @@
+body {
```
MatthewPiatetsky: perhaps when we're done we could run a css linter that could help find the unused rules

public/static-checkout/index.html
```diff
@@ -0,0 +1,148 @@
...
+      </button>
+      <div class="dropdown-menu">
+        <a class="dropdown-item" href="https://courses.edx.org/dashboard">Dashboard</a>
+        <a class="dropdown-item" href="https://courses.edx.org/logout">Logout</a></div>
+    </div>
+  </header>
+  <main>
+    <div class="container-fluid payment-container" aria-label="Payment">
+      <div class="row">
+        <!-- Cart  -->
```
MatthewPiatetsky: we should remove all the comments from the html

public/static-checkout/index.html
```diff
@@ -0,0 +1,148 @@
...
+                <button class="active-button"><img src="assets/accepted-card-logos.png" alt="card-logos"></button>
+                <button><img src="assets/paypal-logo.png" alt="paypal-logo"></button>
+              </div>
+            </div>
+            <!-- Checout Form  -->
+            <div class="card">
+              <div class="checkout-section">
+                <h5>Card Holder Infomation</h5>
+                  <form>
+                    <div class="col-left">
```
MatthewPiatetsky: perhaps these classes could be changed to cl and cr since they reoccur to save space?

public/static-checkout/index.html
```diff
@@ -0,0 +1,148 @@
...
+                    <div class="col-right">
+                      Suite/Apartment Number<br>
+                      <input type="text" name="address2"><br>
+                    </div>
+
+                    <div class="col-left">
+                      City (required)<br>
+                      <input type="text" name="city"><br>
+                    </div>
+                    <div class="col-right">
```
MatthewPiatetsky: country/expiration month and expiration year are dropdowns on the mfe.
i think making at least the expiration fields dropdown here would be a good idea if a wrong format month/year would cause the payment to fail.  
Not sure about country, that's more important for the sdn check than the payment itself, so probably a user inputted country is fine (though we'll get some different spellings)

public/static-checkout/css/style.css
MatthewPiatetsky: this should be minified but only when we're done

public/static-checkout/index.html
```diff
@@ -0,0 +1,148 @@
+<!DOCTYPE html>
```
MatthewPiatetsky: this should be minified but only when we're done

public/static-checkout/index.html
julianajlk: Good call, I totally forgot to have them as dropdowns - updated on the next commit

public/static-checkout/css/style.css
```diff
@@ -0,0 +1,256 @@
...
+
+.dropdown-item:hover {
+  background-color: #f9f9f9;
+}
+
+.container-fluid {
+  margin: auto;
+  max-width: 1440px;
+  width: 100%;
+  box-sizing: border-box;
```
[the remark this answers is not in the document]
julianajlk: removed on next commit

public/static-checkout/css/style.css
```diff
@@ -0,0 +1,256 @@
...
+  color: #171c29;
+}
+
+*, :after, :before {
+  box-sizing: border-box;
+  -webkit-box-sizing: border-box;
+  -moz-box-sizing: border-box;
+}
+
+a:visited {
```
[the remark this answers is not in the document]
julianajlk: Updated!

public/static-checkout/index.html
julianajlk: I've changed the CSS to use the .col classes I had created instead of floating.

PR #359, opened 2020-02-24
Minify svgs, remove some unused css, fix country and state selectors, fix user menu ui css and make coupon box create a new basket with the coupon applied

MatthewPiatetsky:
addressing a few of my comments on this PR
https://github.com/edx/frontend-app-payment/pull/358


public/static-checkout/css/style.css
```diff
@@ -210,11 +192,6 @@ header {
   transform: translate(0%, 25%);
 }
 
-.grid-container {
-  width: 100%;
-  max-width: 1380px;
```
julianajlk: How does it look on larger than laptop screen size?

public/static-checkout/css/style.css
```diff
@@ -227,30 +204,14 @@ header {
...
-  display: table ;
-  clear:both;
-}
-
 [class*='col-'] {
   display: inline-block;
   min-height: 1px;
 }
 
-.col-1 {width: 8.33%;}
```
julianajlk: I was going to use some of them for the responsiveness, but then I would need to add the class to the HTML, so I will just use width + media queries instead

public/static-checkout/assets/avatar.svg
```diff
@@ -1,4 +1 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-  <path d="M4.10255106,18.1351061 C4.7170266,16.0581859 8.01891846,14.4720277 12,14.4720277 C15.9810815,14.4720277 19.2829734,16.0581859 19.8974489,18.1351061 C21.215206,16.4412566 22,14.3122775 22,12 C22,6.4771525 17.5228475,2 12,2 C6.4771525,2 2,6.4771525 2,12 C2,14.3122775 2.78479405,16.4412566 4.10255106,18.1351061 Z M12,24 C5.372583,24 0,18.627417 0,12 C0,5.372583 5.372583,0 12,0 C18.627417,0 24,5.372583 24,12 C24,18.627417 18.627417,24 12,24 Z M12,13 C9.790861,13 8,11.209139 8,9 C8,6.790861 9.790861,5 12,5 C14.209139,5 16,6.790861 16,9 C16,11.209139 14.209139,13 12,13 Z" fill="#686b73"/>
-</svg>
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24"><path d="M4.103 18.135c.614-2.077 3.916-3.663 7.897-3.663s7.283 1.586 7.897 3.663C21.215 16.44 22 14.312 22 12c0-5.523-4.477-10-10-10S2 6.477 2 12c0 2.312.785 4.44 2.103 6.135zM12 24C5.373 24 0 18.627 0 12S5.373 0 12 0s12 5.373 12 12-5.373 12-12 12zm0-11a4 4 0 1 1 0-8 4 4 0 1 1 0 8z" fill="#686b73"/></svg>
```
julianajlk: Thanks for minifying this!

public/static-checkout/js/scripts.js
```diff
@@ -250,27 +239,39 @@ const countryList = {
...
-  VI: 'Virgin Islands (U.S.)',
+  VG: 'Virgin Islands, British',
+  VI: 'Virgin Islands, U.S.',
   WF: 'Wallis and Futuna',
   EH: 'Western Sahara',
   YE: 'Yemen',
   ZM: 'Zambia',
   ZW: 'Zimbabwe',
   AX: 'Åland Islands',
+  BQ: 'Bonaire, Sint Eustatius and Saba',
```
julianajlk: I'm surprised the list I got was so out of date. Let's do this in alphabetical order?

public/static-checkout/css/style.css
julianajlk: Also the .row class keeps the container inside the padding (bottom now has no padding). You're correct that the :before :after is not used right now, but we can keep the .row or add margin-bottom to the .container-fluid and remove the <div class="row"> on the HTML

PR #360, opened 2020-03-04
Static Checkout Page for REV A/B Test

julianajlk:
[REV-1111](https://openedx.atlassian.net/browse/REV-1111). 
This is not to be used outside of the purpose of the test, Revenue team is conducting a static checkout A/B test to determine the fastest loading time it would take to fully load the page. 

All styling is done via custom CSS. There is one visual change from the Payment MFE in the navigation menu, we're not displaying the username, but just a user avatar icon. 

Not included in this test: 
- Auto applied discounts. Coupons are being redirected to the Payment MFE.
- Bundle purchases.
- Form validation. 
- i18n. 
- Local currency. 
- Optimizely. 
- Logging. 
- Experiments. 
- Google Tag Manager. 
- Google Analytics. 
- New Relic. 
- Sailthrough email tracking. 

We're using ES5, and CSS and JS files will be later minified and added to the HTML file.


public/static-checkout/index.html
```diff
@@ -0,0 +1,435 @@
...
+                      <option>2032</option>
+                      <option>2033</option>
+                      <option>2034</option>
+                      <option>2035</option>
+                    </select>
+                  </div>
+                </div>
+              </div>
+
+              <button id="place-order-button" class="submit-button">Place Order</button>
```
abutterworth: It would be better/safer to make this a `<button type="submit">` and handle the form submit event instead of the click event on this button done below: `checkoutButton.addEventListener('click', function submitCybersource(event)...`. So instead:
```
<form id="checkout-form">...</form>
```
```
checkoutForm.addEventListener('submit', submitCybersource...)
```

public/static-checkout/index.html
```diff
@@ -0,0 +1,435 @@
...
+            </form>
+          </div>
+        </section>
+      </div>
+    </div>
+  </main>
+  <script src="js/scripts.js"></script>
```
abutterworth: You might consider putting the content of scripts.js inline rather than having the browser fetch another file.

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,309 @@
...
+}
+
+document.addEventListener('DOMContentLoaded', function handleCountrySelection() {
+  renderCountryState();
+});
+
+couponButton.addEventListener('click', function couponRedirectToMicrofrontend(event) {
+  event.preventDefault();
+  var couponCode = document.getElementById('couponcode').value;
+  redirectToMFE(couponCode);
```
davidjoy: Is redirecting to the MFE for carts that use coupons going to skew the results?  Seems like all those carts would have lower value on average.

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,309 @@
...
+  var urlParams = new URLSearchParams(queryString);
+  var sku = urlParams.get('sku');
+  var newUrl = ecommerceBaseUrl + '/basket/add/?sku=' + sku;
+  if (couponCode) {
+    newUrl = newUrl.concat('&coupon=', couponCode);
+  }
+  window.location.href = newUrl;
+}
+
+userButton.addEventListener('click', function toggleUserDropdown() {
```
davidjoy: The callback function here can just be anonymous -

```
userButton.addEventListener('click', function() {
  document.querySelector('.dropdown-menu').classList.toggle('show');
}, false);
```

The function name isn't really used or necessary.

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,309 @@
...
+  var urlParams = new URLSearchParams(queryString);
+  // Emma: sql or other injection?
+  var basketId = urlParams.get('basketId');
+  // Emma: what if the basketId is not in the param?
+  return basketId;
+}
+
+function redirectToMFE(couponCode) {
+  var queryString = window.location.search;
+  var urlParams = new URLSearchParams(queryString);
```
abutterworth: Heads up `URLSearchParams` isn't available in IE11

public/static-checkout/js/scripts.js
davidjoy: The same goes elsewhere - anywhere that you're passing a function in as a callback or assigning it to an event handler,  (like `window.onclick =` below), doesn't need a name.

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,309 @@
...
+  // Emma: sql or other injection?
+  var basketId = urlParams.get('basketId');
+  // Emma: what if the basketId is not in the param?
+  return basketId;
+}
+
+function redirectToMFE(couponCode) {
+  var queryString = window.location.search;
```
davidjoy: Any variable like `queryString` here that's only being used in one following statement isn't really buying you much but additional bytes.

```
var urlParams = new URLSearchParams(window.location.search);
```

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,309 @@
...
+
+userButton.addEventListener('click', function toggleUserDropdown() {
+  document.querySelector('.dropdown-menu').classList.toggle('show');
+}, false);
+
+window.onclick = function closeMenu(event) {
+  if (!event.target.matches('.user-button')) {
+    var dropdownMenu = document.getElementsByClassName('dropdown-menu');
+    var i;
+    for (i = 0; i < dropdownMenu.length; i += 1) {
```
davidjoy: Think this and line 121 above could just be:

```
for (var i = 0; i < dropdownMenu.length; i++) {
```

public/static-checkout/js/scripts.js
davidjoy: (and there are some long names here and there!)

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,309 @@
...
+  event.preventDefault();
+  var couponCode = document.getElementById('couponcode').value;
+  redirectToMFE(couponCode);
+}, false);
+
+
+function parseCookies() {
+  var cookies = {};
+  if (document.cookie && document.cookie !== '') {
+    document.cookie.split(';').forEach(function (c) {
```
davidjoy: I can't remember what browser support for `forEach` is.  You may want to just use a standard `for` loop here - it'd probably be less verbose than having the anonymous function and all that.

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,309 @@
...
+        cookies[m[1]] = decodeURIComponent(m[2]);
+      }
+    });
+  }
+  return cookies;
+}
+
+function disableForm() {
+  var toDisable = document.querySelectorAll('input, select, button');
+  Array.prototype.forEach.call(toDisable, function (elementToDisable) {
```
davidjoy: Again, may wanna watch the `forEach` - it's also generally slower than a normal for loop, I think.

public/static-checkout/js/scripts.js
davidjoy: Eh, looks like forEach is pretty safe.

public/static-checkout/index.html
julianajlk: Adding both JS and CSS files inline, thanks!

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,309 @@
...
+        form.appendChild(cardExpirationElement);
+
+        document.body.appendChild(form);
+        form.submit();
+      } else {
+        redirectToMFE();
+      }
+    }
+  };
+  var urlEncodedDataPairs = [];
```
davidjoy: Pretty sure this is significantly shorter but equivalent:

```
function encoder(name) {
  returnencodeURIComponent(document.getElementById(name).value)
}
var urlEncodedDataPairs = [
  'basket=' + encodeURIComponent(getBasketId()),
  'first_name=' + encoder('firstname'),
  'last_name=' + encoder('lastname'),
  'address_line1=' + encoder('address1'),
  'address_line2=' + encoder('address2'),
  'city=' + encoder('city'),
  'country=' + encoder('country'),
  'state=' + encoder('state'),
  'postal_code=' + encoder('zip'),
];
```

URL encoding the keys isn't doing anything - the underscore is a valid character in a URL, and the rest is alphanumeric.  I haven't tested this code.

public/static-checkout/js/scripts.js
julianajlk: @davidjoy This was an ESList error fix, https://eslint.org/docs/rules/no-plusplus. Is disabling this the best option?

julianajlk: Removed this rule in @frontend-build eslinter

public/static-checkout/index.html
```diff
@@ -0,0 +1,435 @@
...
+          <p>Add coupon code (optional)</p>
+          <div style="display: inline-flex; margin-bottom: 1rem;">
+            <input id="couponcode" style="margin-right: 0.5rem; width: 12rem;" type="text" name="couponcode" aria-label="coupon code">
+            <button id="coupon-button" class="submit-button" style="font-size: 1rem; padding: 0.5rem 1.5rem;">Apply</button>
+          </div>
+          <h6 class="col-10 col-nums" aria-level="3">TOTAL</h6>
+          <span id="total"><strong>$350.00</strong></span>
+        </div>
+        <div>
+          <h5 aria-level="2">Order Details</h5>
```
MatthewPiatetsky: the spacing here looks a bit off/is different from payment.edx.org

public/static-checkout/index.html
```diff
@@ -0,0 +1,435 @@
...
+                      <option value="VN">Viet Nam</option>
+                      <option value="VG">Virgin Islands, British</option>
+                      <option value="VI">Virgin Islands, U.S.</option>
+                      <option value="WF">Wallis and Futuna</option>
+                      <option value="EH">Western Sahara</option>
+                      <option value="YE">Yemen</option>
+                      <option value="ZM">Zambia</option>
+                      <option value="ZW">Zimbabwe</option>
+                      <option value="AX">Åland Islands</option>
+                      <option value="BQ">Bonaire, Sint Eustatius and Saba</option>
```
MatthewPiatetsky: spelling error

public/static-checkout/index.html
```diff
@@ -0,0 +1,435 @@
...
+                      <option value="CW">Curaçao</option>
+                      <option value="GG">Guernsey</option>
+                      <option value="IM">Isle of Man</option>
+                      <option value="JE">Jersey</option>
+                      <option value="ME">Montenegro</option>
+                      <option value="BL">Saint Barthélemy</option>
+                      <option value="MF">Saint Martin (French part)</option>
+                      <option value="RS">Serbia</option>
+                      <option value="SX">Sint Maarten (Dutch part)</option>
```
MatthewPiatetsky: spelling error 
it seems like we shouldn't need to enter any of these manually, i believe i included somewhere a list that i just got directly from the production drop down

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,313 @@
...
+  if (couponCode) {
+    newUrl = newUrl.concat('&coupon=', couponCode);
+  }
+  window.location.href = newUrl;
+}
+
+function getBasketId() {
+  var url = window.location.href;
+  var basketParams = /[?&]basketId=([^&#]*)/.exec(url);
+  if (!basketParams) {
```
MatthewPiatetsky: do we expect this to happen or is this just defensive coding?

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,313 @@
...
+    }
+  };
+  var basketId = getBasketId();
+  xhr.send(JSON.stringify({
+    basketId,
+    payment_processor: 'paypal',
+  }));
+});
+
+checkoutForm.addEventListener('submit', function submitCybersource(event) {
```
MatthewPiatetsky: is this just copying what the MFE is doing? this code seems rather complex so i would have more questions if it is not

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,313 @@
+/* eslint-disable no-plusplus */
```
MatthewPiatetsky: is eslint in travis or just the precommit hook? if it is the latter, we could remove all the eslint disable lines and commit with the --no-verify option, would save a few lines

public/static-checkout/js/scripts.js
julianajlk: Just added for in case there is so basket id in the url - defensive coding

public/static-checkout/js/scripts.js
julianajlk: The "no i++" was disabled in https://github.com/edx/eslint-config/pull/6. The others I am not sure, but with --no-verify we would still fail in travis

public/static-checkout/js/scripts.js
julianajlk: Emma can better answer this, but we can discuss on slack

emma-green: It is. :)

public/static-checkout/js/scripts.js
```diff
@@ -0,0 +1,313 @@
...
+    for (var i = 0; i < dropdownMenu.length; i++) {
+      var open = dropdownMenu[i];
+      if (open.classList.contains('show')) {
+        open.classList.remove('show');
+      }
+    }
+  }
+};
+
+function redirectToMFE(couponCode) {
```
MatthewPiatetsky: I imagine we will want to exclude learners who get redirected to the MFE from the experiment and we will want to exclude them from the routing as well, so we probably want to add a query parameter here to reflect that we are coming from the static checkout page for the routing code to look at.

public/static-checkout/index.html
```diff
@@ -0,0 +1,435 @@
...
+                <h6 aria-level="3">Title HERE</h6>
+                <p>Verified Certificate</p>
+              </div>
+            </div>
+          </div>
+        </section>
+        <div role="region">
+          <h5 aria-level="2">Summary</h5>
+          <span class="col-10 col-nums">Price</span>
+          <span id="summaryprice">$350.00</span>
```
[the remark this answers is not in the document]
julianajlk: I thought fixed that - I don't see this happening on my localhost. What width?

public/static-checkout/index.html
julianajlk: What should it be? On Wikipedia it's the same.

public/static-checkout/index.html
julianajlk: Spelling error - you mean Sint should be Saint? 
Emma decided it would be better to have them in the HTML already then how I did it before in the JS file as an object with a for loop.

public/static-checkout/index.html
```diff
@@ -0,0 +1,435 @@
...
+      <nav aria-label="Main">
+        <button class="user-button" aria-label="Account menu for user" aria-haspopup="menu" aria-expanded="true">
```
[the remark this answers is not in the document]
julianajlk: I brought this up in the beginning when I started, folks seemed to like the idea of not having the gray button but we can rethink. Maybe as UX?

public/static-checkout/index.html
MatthewPiatetsky: yup
agree they should be in the html, i just meant we don't have to do any of the html by hand, which could result the spelling errors

public/static-checkout/index.html
MatthewPiatetsky: oh i assumed it was meant to be saint, nvm. (although the spelling on payment.edx.org is completely different) i feel like just using identical spelling to payment.edx.org might be ideal

public/static-checkout/index.html
MatthewPiatetsky: sorry i incorrectly assumed Sint was supposed to be Saint (though i do think just programmatically copying the spellings from the dropdown on payment.edx.org might be ideal since it is different)

public/static-checkout/index.html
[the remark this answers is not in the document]
MatthewPiatetsky: impersonating iphone 6/7/8 in the device impersonating section on chrome

public/static-checkout/index.html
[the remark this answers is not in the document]
MatthewPiatetsky: not my decision, was just pointing out how I saw it, but if emma or shelby think this is fine then it is fine

public/static-checkout/js/scripts.js
emma-green: hmm... @inventhouse would we know this information at analysis time because we see them on both pages?

MatthewPiatetsky: queries like that aren't trivial to write in the analysis, but regardless we will need a parameter to exclude those users from the routing, otherwise they could get routed right back to the static checkout page

public/static-checkout/index.html
julianajlk: It's the same list as in payment.edx.org

PR #361, opened 2020-03-11
Ensure we are able to checkout with cybersource and load the paypal payment page

MatthewPiatetsky:
The routing code in ecommerce was assuming a certain structure. The structure was different here so I was making changes to fix that.
I also changed what we are sending to cybersource to fix some errors I ran into when trying to check out.
Paypal checkout still didn't work for me but I'm not sure if it works locally, was hoping to test on stage or eventually prod

### DIFF
--- a/public/static-checkout/js/scripts.js
+++ b/public/static-checkout/js/scripts.js
@@ -314,4 +314,3 @@ checkoutForm.addEventListener('submit', function submitCybersource(event) {
   var urlEncodedData = urlEncodedDataPairs.join('&').replace(/%20/g, '+');
   xhr.send(urlEncodedData);
 });
-
--- a/public/static-checkout/js/scripts.js
+++ b/public/static-checkout/js/scripts.js
@@ -3,6 +3,9 @@
 /* eslint-disable prefer-template */
 /* eslint-disable prefer-arrow-callback */
 /* eslint-disable no-var */
+/* eslint-disable camelcase */
+/* eslint-disable func-names */
+/* eslint-disable prefer-destructuring */
 var checkoutForm = document.getElementById('checkout-form');
 var paypalButton = document.getElementById('submit-paypal');
 var couponButton = document.getElementById('coupon-button');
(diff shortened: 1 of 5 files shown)


public/static-checkout/js/scripts.js
```diff
@@ -3,84 +3,87 @@
...
-    UT: 'Utah',
-    VT: 'Vermont',
-    VA: 'Virginia',
-    WA: 'Washington',
-    WV: 'West Virginia',
-    WI: 'Wisconsin',
-    WY: 'Wyoming',
-  },
+  CA: [
+    ['AB', 'Alberta'],
```
MatthewPiatetsky: need to send the short codes to cybersource on checkout

public/static-checkout/js/scripts.js
```diff
@@ -215,9 +219,9 @@ paypalButton.addEventListener('click', function submitPaypal() {
       }
     }
   };
-  var basketId = getBasketId();
+  var basket_id = getBasketId();
   xhr.send(JSON.stringify({
-    basketId,
+    basket_id,
```
MatthewPiatetsky: ecommerce expects basket id in snake case

webpack.prod.config.js
```diff
@@ -34,7 +34,7 @@ config.plugins = [
   // Stage also uses this config and dev uses this config when running npm run build
   new CopyPlugin([
     {
-      from: path.resolve(__dirname, 'public/REV1074'),
+      from: path.resolve(__dirname, 'public/static-checkout'),
```
MatthewPiatetsky: this is what pulls in the file into the dist folder (necessary to access it on prod)

public/static-checkout/js/scripts.js
```diff
@@ -108,8 +111,8 @@ window.onclick = function closeMenu(event) {
 };
 
 function redirectToMFE(couponCode) {
-  var url = window.location.href;
-  var sku = /static-checkout\/([^&#]*).html/.exec(url)[1];
+  var url = window.location.pathname;
+  var sku = /.*\/([^&#]*).html/.exec(url)[1];
```
MatthewPiatetsky: this regex wasn't working for me

public/static-checkout/js/scripts.js
```diff
@@ -118,8 +121,8 @@ function redirectToMFE(couponCode) {
 }
 
 function getBasketId() {
-  var url = window.location.href;
-  var basketParams = /[?&]basketId=([^&#]*)/.exec(url);
+  var url = window.location.search;
+  var basketParams = /[?&]basket_id=([^&#]*)/.exec(url);
```
MatthewPiatetsky: this regex wasn't working for me

public/static-checkout/js/scripts.js
julianajlk: Why wasn't it working? I'm curious because I get the sku when I try on console

MatthewPiatetsky: not sure, i get that now too but pretty sure at one point it didn't work

PR #362, opened 2020-03-19
Add whitelist with certain npm packages with vulnerabilities

julianajlk:
Upon `npm audit` there are certain package vulnerabilities that do not have a quick fix at the moment - this should not stop us from having commits and merging PRs. 

By utilizing [audit-ci](https://www.npmjs.com/package/audit-ci), we can whitelist these specific packages. Now the pre-commit hooks will run the whitelisted command instead of `npm audit`.
- [minimist](https://www.npmjs.com/advisories/1179) (low). 
- [acorn](https://npmjs.com/advisories/1488) (moderate): current version of es-check accepts affected acorn version 6.0.4, latest version also accepts affected acorn version 6.1.1. 
- [decompress](https://npmjs.com/advisories/1217) (high): no fix is currently available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,13 @@
 language: node_js
 node_js: 12
 before_install:
-- npm install -g npm@6
+- npx audit-ci -m -w acorn decompress minimist
 install:
 - npm ci
 script:
 - make validate-no-uncommitted-package-lock-changes
 - npm run i18n_extract
 - npm run lint
-- npm audit
 - npm run test
 - npm run build
 - npm run is-es5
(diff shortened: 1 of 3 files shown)


julianajlk:
> if we whitelist these does that mean we won't realize when there is a fix and so it won't get fixed?

We can still run `npm audit` to see vulnerabilities, but now vulnerabilities in these 3 packages are not blocking.

quinnturner:
@MatthewPiatetsky author of `audit-ci` here. When you whitelist a package or advisory that does not have a vulnerability, `audit-ci` should `console.warn` in yellow to outline that the whitelisting was unnecessary. When this happens, it's OK to remove from the whitelist. Periodically checking the logs (or running `npm audit` as @julianajlk described) is the appropriate course of action.

PR #364, opened 2020-03-23
[REV-1126] Add a script to generate both the minified html for a generic static page and the pages themselves.

MatthewPiatetsky:
Also adds the actual pages we want to use for the test


public/static-checkout/generate_static_pages.sh
```diff
@@ -0,0 +1,21 @@
...
+JS_FILE_NAME="unminified.js"
+html-minifier --collapse-whitespace $HTML_FILE_NAME -o dist/minified.html
+echo "$HTML_FILE_NAME minified"
+cssnano $CSS_FILE_NAME dist/minified.css
+echo "$CSS_FILE_NAME minified"
+uglifyjs --compress --mangle -o dist/minified.js -- $JS_FILE_NAME
+echo "$JS_FILE_NAME minified"
+echo "Minification complete!"
+python3 generate_static_pages.py
+echo "Pages generated!"
```
inventhouse: Nit: if you make any changes to this, add a newline to the end

PR #365, opened 2020-03-24
Added events using the same mechanism as static checkout pages

inventhouse:
Added some test events using the same mechanism we intend to use for REV-1074 static checkout pages

Note that I wrote the `sendRev1074Event` with an eye to being able to just drop it into the static pages when we add events there.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -21,7 +21,7 @@ import Header, { messages as headerMessages } from '@edx/frontend-component-head
 import Footer, { messages as footerMessages } from '@edx/frontend-component-footer';
 
 import appMessages from './i18n';
-import { PaymentPage, EcommerceRedirect, responseInterceptor, markPerformanceIfAble, getPerformanceProperties } from './payment';
+import { PaymentPage, EcommerceRedirect, responseInterceptor, markPerformanceIfAble, getPerformanceProperties, sendRev1074Event } from './payment';
 import configureStore from './data/configureStore';
 
 import './index.scss';
@@ -49,6 +49,8 @@ subscribe(APP_READY, () => {
     'edx.bi.ecommerce.payment_mfe.started_painting',
     getPerformanceProperties(),
   );
+  sendRev1074Event('payment_mfe.started_painting', {});
+
   ReactDOM.render(
     <AppProvider store={configureStore()}>
       <Header />
--- a/src/payment/cart/OrderSummary.jsx
+++ b/src/payment/cart/OrderSummary.jsx
@@ -14,7 +14,9 @@ class OrderSummary extends React.Component {
       getPerformanceProperties(),
     );
     let eventData = {};
-    let renderTime = window.performance.getEntriesByName("Order Summary component rendered");
+    let renderTime = window.performance &&
+      window.performance.getEntriesByName &&
+      window.performance.getEntriesByName("Order Summary component rendered");
     if (renderTime) {
       eventData["summaryRenderTiming"] = renderTime[renderTime.length - 1].toJSON();
     }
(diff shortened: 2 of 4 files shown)


src/payment/cart/OrderSummary.jsx
```diff
@@ -12,6 +13,12 @@ class OrderSummary extends React.Component {
       'edx.bi.ecommerce.payment_mfe.order_summary_rendered',
       getPerformanceProperties(),
     );
+    let eventData = {};
+    let renderTime = window.performance.getEntriesByName("Order Summary component rendered");
+    if (renderTime) {
+      eventData["summaryRenderTiming"] = renderTime[renderTime.length - 1].toJSON();
```
MatthewPiatetsky: what is this?

src/index.jsx
```diff
@@ -49,6 +49,8 @@ subscribe(APP_READY, () => {
     'edx.bi.ecommerce.payment_mfe.started_painting',
     getPerformanceProperties(),
   );
+  sendRev1074Event('payment_mfe.started_painting', {});
```
MatthewPiatetsky: you're adding all of these after sendTrackEvent but doesn't that function go to the same api?

src/payment/cart/OrderSummary.jsx
inventhouse: That gets the timing mark for when the Order Summary component finishes rendering (if available; taking the last one if multiple, which there shouldn't be, but `getEntriesByName` returns an array)

src/payment/index.js
```diff
@@ -1,3 +1,33 @@
...
+/* eslint-disable prefer-template */
+/* eslint-disable dot-notation */
+export function sendRev1074Event(eventType, eventData) {
+  eventData['_export'] = 'false'; // Don't let these events be exported to partners
+  if (window.performance.timing) {
+    eventData.timing = window.performance.timing.toJSON();
+  }
+  const encodedEvent = [
+    'event_type=edx.experiment.rev1074.' + encodeURIComponent(eventType),
+    'page=' + encodeURIComponent(window.location.href),
```
MatthewPiatetsky: why do we need to encode the different variables? is this for security or something

src/payment/cart/OrderSummary.jsx
MatthewPiatetsky: so what happens if renderTime is not defined? do we not have a time? Should the sendEvent be inside this if block then?

src/payment/index.js
inventhouse: As far as I could tell, the GET does _not_ encode them automatically, please let me know if there's a better way.  (Note: this is just encoding, not encrypting, no security implications around this)

src/payment/cart/OrderSummary.jsx
inventhouse: If we don't have a `renderTime`, oh well, but we should still send the event anyway so we can compare with the regular track event.

src/index.jsx
inventhouse: No, I believe the regular track events are going to Segment.

src/payment/index.js
MatthewPiatetsky: why is the encoding necessary in that case? if not necessary we could save some characters on the static page

inventhouse: Ooohhh, you mean _that_ specific instance; probably not?  (The "spec" is _very_ unclear about what needs to be encoded in different cases)  I'll try without encoding that one 👍

src/index.jsx
inventhouse: Mainly they're going to different destinations; by adding "duplicate" events going to the API we plan to use for the Static Checkout Page test we can validate that mechanism and try out analyzing these events.

src/payment/index.js
inventhouse: Dropped the encode calls on things that don't really need it.

PR #366, opened 2020-03-24
feat: upgrade frontend build

abutterworth:


abutterworth:
Build passes: https://travis-ci.org/github/edx/frontend-app-payment/builds/666863483?utm_medium=notification&utm_source=github_status

src/payment/cart/Cart.jsx
```diff
@@ -56,72 +56,72 @@ class Cart extends React.Component {
...
-          {products && products.length ? products.map(product => (
+          {products.map(product => (
             <ProductLineItem
               key={product.title}
               isNumEnrolledExperiment={isNumEnrolledExperiment}
               REV1045Experiment={REV1045Experiment}
               isPaymentVisualExperiment={isPaymentVisualExperiment}
               enrollmentCountData={enrollmentCountData}
               {...product}
-            />)) : (REV1045Experiment ?
```
abutterworth: Removing this experiment per @julianajlk feedback. Nested ternary operators violate new lint rules. So it was refactor or delete. Chose to delete

webpack.prod.config.js
```diff
@@ -93,4 +94,15 @@ config.plugins = [
...
+/**
+ * The loader in the default config references a different version of
+ * MiniCssExtractPlugin and will break if we don't swap it with the one used in
+ * this file.
+ *
+ * TODO: This is a very brittle. We should improve this in the future by either
+ * defining the whole of the webpack config here or impementing some better
+ * configuration modification functionality in frontend-build.
+ */
+config.module.rules[2].use[0] = MiniCssExtractPlugin.loader;
```
abutterworth: The most objectionable part of this file. The loader in the default config references a different version of MiniCssExtractPlugin and will break if we don't swap it with the one used in this file.

src/payment/cart/Cart.jsx
```diff
@@ -56,72 +56,72 @@ class Cart extends React.Component {
...
-
-            {showCouponForm ?
-              <CouponForm isPaymentVisualExperiment={isPaymentVisualExperiment} /> : null}
-
-            <TotalTable total={orderTotal} />
-
-            {isCurrencyConverted ? <CurrencyDisclaimer /> : null}
-          </OrderSummary> : (
-            <React.Fragment>
+        {!REV1045Experiment || loaded
```
julianajlk: Could also remove all the places where `REV1045Experiment` is in the code - if not on this PR, REV squad will do in a future experiment clean up.

src/feedback/AlertList.jsx
```diff
@@ -11,7 +11,7 @@ class AlertList extends Component {
   getUserMessage(code, userMessage) {
     if (this.props.messageCodes[code]) {
       return this.props.messageCodes[code];
-    } else if (code === 'fallback-error') {
+    } if (code === 'fallback-error') {
```
davidjoy, line 14: The linter doesn't complain about having this second `if` on the same line? O_o

abutterworth: I guess not... fixing

webpack.prod.config.js
davidjoy: It might be worth rebuilding the rules tree from scratch here, rather than relying on indices, yeah.  If we touch it in frontend-build, it's pretty likely to bust

abutterworth: True, though then all those dev deps need to be managed by this repo. I'd prefer us to have a better way to manipulate configs. In the meantime I'm not sure what the right way to go is

package.json
```diff
@@ -73,30 +73,25 @@
     "universal-cookie": "4.0.2"
   },
   "devDependencies": {
-    "@edx/frontend-build": "^2.0.6",
-    "audit-ci": "^2.5.1",
+    "@edx/frontend-build": "5.0.3",
+    "audit-ci": "2.5.1",
     "axios-mock-adapter": "1.17.0",
     "chance": "1.1.0",
-    "clean-webpack-plugin": "3.0.0",
```
abutterworth, line 80: Leveraging webpack-merge allows us to eliminate re-adding plugins that already exist in our webpack config. presets and therefore this repo doesn't need to manage their versions.

webpack.prod.config.js
abutterworth: Updated with webpack-merge

PR #369, opened 2020-04-01
Add proof-of-concept telemetry event

inventhouse:
Add proof-of-concept tracking event
Also make it easier to `generate_static_pages` that are _not_ minified for debugging

Please review:
>  M public/static-checkout/generate_static_pages.py
>  M public/static-checkout/generate_static_pages.sh
>  M public/static-checkout/unminified.js


public/static-checkout/generate_static_pages.py
```diff
@@ -1,17 +1,17 @@
 import csv
 # Combine minified html/css/js into one string
 with open('dist/minified.html', 'r') as file:
-    html = file.read().replace('\n', '')
```
MatthewPiatetsky, line 4: why was this removed?

inventhouse: They appear to be unnecessary in the minified case (based on my testing) and they interfere with generating non-minified pages for debugging.

public/static-checkout/unminified.js
```diff
@@ -169,6 +200,12 @@ document.addEventListener('DOMContentLoaded', function () {
   renderCountryState();
 });
 
+/* eslint-disable no-unused-vars */
+window.onload = function sendPageLoaded(event) {
+  sendRev1074Event('static.page_loaded', {});
```
MatthewPiatetsky, line 205: is this the only event we need?

inventhouse: No, there is at least a couple of others (see REV-1116)

PR #375, opened 2020-04-07
[REV-1140] Add static page for stage course for testing on stage and fix bug in the script

MatthewPiatetsky:


public/static-checkout/generate_static_pages.py
```diff
@@ -15,9 +15,9 @@
 
 # Read in course data and generate a page for each course
 with open('courselist.csv', newline='') as csvfile:
-    coursehtml = (html + '.')[:-1]
     csvreader = csv.reader(csvfile, delimiter=',')
     for row in csvreader:
+        coursehtml = (html + '.')[:-1]
```
MatthewPiatetsky, line 20: the deepcopy of the html  was happening outside the loop, so the html from the first course was being reused

julianajlk: Great catch!

PR #376, opened 2020-04-07
Fix credit card logos and Paypal png 

julianajlk:
[REV-1158](https://openedx.atlassian.net/browse/REV-1158). 
- Added PNG files for credit card and Paypal logos to assets, updated CopyPlugin to copy assets folder as well. 
- Removed original HTML page as this will no longer be updated when we run `generate_static_pages.sh`.
- Changed the img src url to absolute url instead of relative url.


julianajlk:
To do: we should delete `public/static-checkout/8580DD0.html` since the Intermediate Happiness course will be added to the csv file

public/static-checkout/unminified.html
```diff
@@ -63,8 +63,8 @@ <h5 aria-level="2">Order Details</h5>
           <div class="mb">
             <h5 aria-level="2">Select Payment Method</h5>
             <div class="payment-method">
-              <button class="active-button"><img src="https://payment.edx.org/159f11b7867975b80646a8626d3ca344.png" alt="card-logos"></button>
-              <button id="submit-paypal"><img src="https://payment.edx.org/e21c8a065b1dbe44064a5828edec39c6.png" alt="paypal-logo"></button>
+              <button class="active-button"><img src="../dist/assets/creditcard.png" alt="card-logos"></button>
```
MatthewPiatetsky, line 66: this might work but i'm not sure.  
the pages being copied over are accessible at payment.edx.org/<sku>.html so what might work is payment.edx.org/creditcard.png.

julianajlk: Interesting, so the directories `pages` and `assets` would not be in the final url - it would not be payment.edx.org/pages/.html or payment.edx.org/assets/creditcard.png

PR #377, opened 2020-04-08
[REV-1140] Use correct CSRF token in requests to ecommerce

MatthewPiatetsky:
I did some testing with a debugger on stage which showed that this is the correct CSRF token, though we won't know if this exact code works until we deploy to stage

You can filter out the html files with the file filter button on the files tab


MatthewPiatetsky:
yes i believe the cookie had the wrong csrf token, making a request to ecommerce to get this token seemed to work with a debugger

PR #379, opened 2020-04-14
Remove call to SDN search endpoint and associated tests 

julianajlk:
[REV-1125](https://openedx.atlassian.net/browse/REV-1125). 
- Removing the call to the SDN search endpoint (api/v2/sdn/search/) because we've added the SDN check to the Cybersource api submit (payment/cybersource/api-submit), so this will remove the duplicated SDN check. 
- Removing related tests.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,7 +53,6 @@ and log in with a staff account. Related endpoints:
   - POST /bff/payment/v0/quantity/
   - POST /bff/payment/v0/vouchers/
   - DELETE /bff/payment/v0/vouchers/{voucherid}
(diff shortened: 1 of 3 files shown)


julianajlk:
> do also need to remove from static page and/or ecommerce front end? maybe we don't care about the latter if applicable
1. For static page, we're not doing SDN check (will be done in ecommerce on cybersource submit endpoint). 
2. We have tickets to remove from ecommerce js - 
https://openedx.atlassian.net/browse/REV-1162
https://openedx.atlassian.net/browse/DEPR-90

PR #385, opened 2020-04-30
[REV-1170] emit error message when submitting an invalid code from the static page

MatthewPiatetsky:
this will handle the code that should be sent by https://github.com/edx/ecommerce/pull/2932

### DIFF
--- a/src/feedback/data/sagas.js
+++ b/src/feedback/data/sagas.js
@@ -38,7 +38,7 @@ export function* handleErrors(e, clearExistingMessages) {
   }
 }
 
-export function* handleMessages(messages, clearExistingMessages) {
+export function* handleMessages(messages, clearExistingMessages, url) {
   // If this doesn't contain anything we understand, bail.
   if (!Array.isArray(messages)) {
     return null;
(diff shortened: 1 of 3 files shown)


src/feedback/data/sagas.js
```diff
@@ -48,6 +48,11 @@ export function* handleMessages(messages, clearExistingMessages) {
     yield put(clearMessages());
   }
 
+  if (window.location.search.indexOf('message') > 0) {
+    const message = new URLSearchParams(window.location.search).get('message');
+    yield put(addMessage('message', `${message}`, {}, MESSAGE_TYPES.ERROR));
```
julianajlk, line 53: Where are you using this code 'message'? Maybe you can use a more descriptive word than just 'message', what type of message? You should also update the sagas.test file.

MatthewPiatetsky: i'm using it here
well it is meant to be generic on purpose, i guess i could say error_message if that's clearer since i did make the message_type error

MatthewPiatetsky:
@julianajlk @emma-green do the changes address your comments?

PR #389, opened 2020-05-11
Fix static-checkout IE browser compatibility SVG issues

julianajlk:
[REV-1168](https://openedx.atlassian.net/browse/REV-1168). 
Added width and height to SVGs to fix Internet Explorer rendering.


MatthewPiatetsky:
the issues i was seeing was with the form fields? does this fix those as well?

julianajlk:
> the issues i was seeing was with the form fields? does this fix those as well?
Yep, the form fields looked funky because of the SVGs, once they are in place the fields go to their normal spots.

PR #391, opened 2020-05-12
static checkout: update payment logos to use https

dianekaplan:
Update to eliminate "Mixed Content" console errors (page is secure, but images are http)


dianekaplan:
> You need to generate the static pages with this change. There's a generate pages .sh file with instructions

ah of course- will have a look!

PR #395, opened 2020-05-14
Update list of courses for static checkout experiment

MatthewPiatetsky:
I still need a PR to update the ecommerce skus

To generate this I was using the following query in vertica
```
SELECT
run.title_override, image, sku, price, run.key, course.key, org.course_partner
FROM
discovery_read_replica.course_metadata_seat as seat
JOIN
discovery_read_replica.course_metadata_courserun as run ON seat.course_run_id = run.id
JOIN
discovery_read_replica.course_metadata_course as course ON course.id = run.course_id
JOIN
business_intelligence.course_master as org on org.course_id = run.key
WHERE
seat.draft=0
AND seat.upgrade_deadline > '2020-06-15'
AND seat.type='verified'
AND run.language_id IN ('en-us','en-gb','en-in','en-au')
AND run.title_override IS NOT NULL
AND LENGTH(run.title_override) > 6
AND org.course_partner NOT IN ('edX', 'TsinghuaX');

```
then removing rows with empty columns


emma-green:
> Thanks for including your query! Maybe add the whitelisted package in a separate commit - but if you're going to squash everything nevermind

^ I would go one step further, and say, have a commit with just the page regeneration, so it's easier to review the stuff you actually changed.

julianajlk:
> > Thanks for including your query! Maybe add the whitelisted package in a separate commit - but if you're going to squash everything nevermind
> 
> ^ I would go one step further, and say, have a commit with just the page regeneration, so it's easier to review the stuff you actually changed.

In some cases - when making visual changes - that's more tricky bc you need to generate the pages in order to see it locally, so you would need to commit your changes, then do another commit for the pages generation, and check locally every single time just to see if your change worked.

MatthewPiatetsky:
you can uncheck .html in the files tab to filter out the static pages

PR #401, opened 2020-07-23
 [REV-1261] Track Names Experiment checkout page changes

MatthewPiatetsky:
changes the values of two fields for the experiment by reading a cookie set in the experiment since optimizely was removed from the page


MatthewPiatetsky:
per description cookie is being set in the experiment (in optimizely) so outside of this code

MatthewPiatetsky:
i think the assumption is this code would be deleted after the experiment so it doesn't matter (although there are like 5 non-cleaned up experiments in the code which is super confusing)

julianajlk:
> i think the assumption is this code would be deleted after the experiment so it doesn't matter (although there are like 5 non-cleaned up experiments in the code which is super confusing)

I think there is a clean up ticket for experiments - or we were going to discuss what the process for inserting experiments would be in an ideal world. We should parking lot that.

PR #405, opened 2020-08-06
WIP: Flex Microform

mulby:


mulby:
Public key is only valid for 15 minutes, this spike does not handle that edge case

mulby:
Get a front-end person that can review this

src/payment/checkout/payment-form/CardDetails.jsx
```diff
@@ -209,7 +213,16 @@ export class CardDetailsComponent extends React.Component {
           />
         </h5>
 
-        <div className="row">
+        {this.props.flexMicroformEnabled && <FlexMicroform
+          captureKeyId={this.props.captureKeyId}
```
mulby, line 217: Change all references to "captureKey" to "publicKey"

src/payment/checkout/payment-form/CardDetails.jsx
```diff
@@ -209,7 +213,16 @@ export class CardDetailsComponent extends React.Component {
           />
         </h5>
 
-        <div className="row">
+        {this.props.flexMicroformEnabled && <FlexMicroform
```
mulby, line 216: The flag should probably have 3 states - loading, true and false

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,97 @@
...
+        }
+        this.onLoad = this.onLoad.bind(this);
+        this.getPaymentToken = this.getPaymentToken.bind(this);
+    }
+
+    componentDidMount() {
+        setPaymentTokenFetch(this.getPaymentToken);
+
+        const script = document.createElement('script');
+        script.src = "https://flex.cybersource.com/cybersource/assets/microform/0.11/flex-microform.min.js";
```
mulby, line 20: Performance implications of this?

src/payment/checkout/payment-form/CardDetails.jsx
```diff
@@ -209,7 +213,16 @@ export class CardDetailsComponent extends React.Component {
...
+          captureKeyId={this.props.captureKeyId}
+          render={
+            ({ microform }) => (
```
cpennington, line 219: I think this may be a mis-application of render props, which seem to be about cross-cutting concerns: https://reactjs.org/docs/render-props.html

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,97 @@
...
+        if (this.state.microform !== null) {
+            return;
+        }
+        if (!this.props.captureKeyId) {
+            return;
+        }
+        if (window['Flex'] === undefined) {
+            return;
+        }
+        const microform = new window.Flex(this.props.captureKeyId).microform({
```
cpennington, line 41: Could be a react hook? https://reactjs.org/docs/hooks-intro.html

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,97 @@
...
+            styles: {
```
inventhouse, line 42: hack: hardcoded style

src/payment/checkout/payment-form/CardDetails.jsx
cpennington: I'm not sure if this can just be `props.children` based?

src/payment/checkout/payment-form/flex-microform/FlexMicroformField.jsx
```diff
@@ -0,0 +1,87 @@
...
+        this.remoteState.disabled = this.props.disabled;
+        field.load(this.element.current);
+        if (this.props.onChange) {
+            field.on('change', this.props.onChange);
+        }
+        this.setState({ field });
+    }
+
+    updateRemoteState() {
+        if (this.props.disabled !== this.remoteState.disabled) {
```
cpennington, line 48: fedx consultant: What's the better way to trigger code when a prop changes?

src/payment/checkout/payment-form/flex-microform/CreditCardNumberField.jsx
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    render() {
+        return (<div className="col-lg-6 form-group">
+            <FlexMicroformField
+                fieldType="number"
+                microform={this.props.microform}
+                disabled={this.props.disabled}
+                onChange={this.onNumberChange}
+                label={<FormattedMessage
```
mulby, line 33: Is this handling of i18n correct?

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+        this.state = {
+            microform: null,
+        }
+        this.onLoad = this.onLoad.bind(this);
+        this.getPaymentToken = this.getPaymentToken.bind(this);
+    }
+
+    componentDidMount() {
+        setPaymentTokenFetch(this.getPaymentToken);
```
mulby, line 17: @davidjoy this can't be the right way to handle this?!?!

src/payment/checkout/payment-form/flex-microform/FlexMicroformField.jsx
mulby: note that when you call `this.state.field.update()` it removes focus from the field, so calling it on every keystroke was preventing me from typing continuously into the field. I had to figure out some way to only call that function when this particular prop changed.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+        const script = document.createElement('script');
```
davidjoy, line 19: One concern here is that we're adding this script every time this component mounts.  Any component above this in the component tree could unmount/re-mount this component at any time, in which case we would re-attach the script to the body, which could have pretty undefined results.  Seems like we'd want to attach this script at a more predictable point in the application lifecycle.  Sometime that we know only happens once.

I think the right thing to do here would be to attach it in the API layer when the basket is returned and keep track of whether it's been added.  That could be done by checking if `window.Flex` exists before appending the script, and if it doesn't, appending it.

Another very different option would be to conditionally add the script to the index.html file via an environment variable.  This would, of course, require coordinating the ecommerce basket's `flexMicroformEnabled` flag with the config of this MFE.  That may not be desirable.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
davidjoy: I'd worry more about whether their script is idempotent.  It probably is, but if it's not, this could create some really crazy/broken behavior if it gets called multiple times (see above)

But yeah, we'll have to wait for their script before showing the UI, presumably... It's only 7.3k though.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
davidjoy: This may be hard to fix for now since we don't have our style constants expressed as JS - just in CSS.  

If we later switch to some CSS-in-JS, keeping this in sync with our theme will be a lot easier.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,97 @@
...
+                    'font-size': '16px',
+                    'font-family': 'Roboto, "Helvetica Neue", Arial, sans-serif',
+                    'color': '#2d323e',
+                },
+                ':disabled': {
+                    'cursor': 'not-allowed'
+                }
+            }
+        });
+        this.setState({microform: microform});
```
davidjoy, line 53: Note that while it's not a requirement, state should generally be serializable data.  The concern is around side effects/internal state changes that may occur in the object not being 'noticed' by React if they don't take place through calls to `setState` or changes in props.  If a piece of data in an object is changed outside React, it won't know the data changed and won't be able to re-render as a result, which can cause undefined behavior.

davidjoy: You may consider setting microform as an instance variable on the component, which would indicate it's not supposed to be considered when re-rendering, but would let you maintain a handle to it.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,97 @@
...
+                    reject(err);
+                } else {
+                    resolve(token);
+                }
+            });
+        });
+    }
+
+    componentDidUpdate() {
+        this.initialize();
```
davidjoy, line 76: It's worth noting that componentDidUpdate will occur anytime the component rerenders, which may be more often than you want to call `initialize`.  

If anything, you may consider using the `componentDidUpdate(prevProps, prevState)` form of the method, which will let you compare captureKeyId to its previous value and choose to initialize _only_ when it has changed.  That way if you add more props to the component, you won't run the risk of blowing away your `microform` instance over and over.

src/payment/checkout/payment-form/flex-microform/FlexMicroformField.jsx
davidjoy: `componentDidUpdate(prevProps, prevState)` - that version of `componentDidUpdate` will let you see past props.  Then you can compare them to the current `this.props` to determine if you should trigger code.

You can also do this by memoizing the function containing the code you want to trigger, if you want to get fancy.  But there's nothing wrong with the above and it's pretty understandable.

Memoization example for the curious: https://github.com/edx/frontend-app-learning/blob/master/src/courseware/CoursewareContainer.jsx#L72 (Note, `lodash.memoize` does NOT work the way you'd expect!)

src/payment/checkout/payment-form/CardDetails.jsx
davidjoy: Yeah, I'm not sure render props is buying you much here.  Might be easier to read to just include the <div> below in the actual <FlexMicroform> component's render function directly.  

You might use render props if you needed to render FlexMicroform with different children in multiple versions of a <CardDetails> component, and those children needed access to a set of props/state defined in FlexMicroform.  The second half of that is what's happening, but we don't seem to have the "different children in multiple places" part.  Lemme know if this isn't making sense.

src/payment/checkout/payment-form/CardDetails.jsx
```diff
@@ -95,6 +98,7 @@ export class CardDetailsComponent extends React.Component {
 
   render() {
     const { disabled, isPaymentVisualExperiment } = this.props;
+    const loading = this.props.captureKeyId === '';
```
davidjoy, line 101: Is this variable being used?  Doesn't seem to be.

src/payment/checkout/payment-form/PaymentForm.jsx
```diff
@@ -219,7 +240,12 @@ export class PaymentFormComponent extends React.Component {
           disabled={disabled}
           isPaymentVisualExperiment={isPaymentVisualExperiment}
         />
-        <CardDetails disabled={disabled} isPaymentVisualExperiment={isPaymentVisualExperiment} />
+        <CardDetails
```
davidjoy, line 243: Looking at how much the code here diverges depending on whether flexMicrofrormEnabled is true or not, it may be worth having two different CardDetails components and move some of the above conditional logic into them.  

When a set of components starts having lots of conditionals like this, it's usually a sign that they're not really one simple component hierarchy, but two separate ones.

I'm not 100% on exactly what that change would look like, but I think it'd make 1) the existing components nearly identical to what they are now, and 2) would put the flex microform code in a component by itself, isolated from the rest of the code that isn't being used in that workflow.

src/payment/checkout/payment-form/flex-microform/CreditCardNumberField.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+import FlexMicroformField from './FlexMicroformField';
+
+class CreditCardNumberField extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            cardIcon: null,
+        }
+        this.onNumberChange = this.onNumberChange.bind(this);
```
davidjoy, line 15: FYI - you can avoid needing to remember `bind` lines like this by declaring the method as:

```
onNumberChange = (data) => {

};
```

The arrow function ensures that `this` is always bound to this scope.

src/payment/checkout/payment-form/flex-microform/CreditCardNumberField.jsx
davidjoy: It's okay.  You're passing a component in and rendering it - it's not super canonical, but it works.  

I think the ideal form would be to use the `this.props.intl.formatMessage()` form of localization here, which will result in a string that you pass in, rather than a component.  It does mean you'll need a messages.js file to hold the messages.  It's better for passing the translated strings around as 'data' like this.

src/payment/checkout/payment-form/flex-microform/FlexMicroformField.jsx
davidjoy: (On second thought, ignore all this memoization stuff - that first bit will do _just fine_)

src/payment/data/selectors.js
```diff
@@ -62,6 +62,7 @@ export const paymentSelector = createSelector(
         basket.loaded && !basket.redirect && (!basket.products || basket.products.length === 0),
       isRedirect:
         (basket.loaded && !!basket.redirect) || (!basket.loaded && isCouponRedeemRedirect),
+      captureKeyId: basket.captureContext ? basket.captureContext.keyId : '',
```
davidjoy, line 65: It's a bit of a nit, but I'd favor `null` here instead of `''` - we don't really use the latter for falsy values very much.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,97 @@
...
+                },
+                ':disabled': {
+                    'cursor': 'not-allowed'
+                }
+            }
+        });
+        this.setState({microform: microform});
+    }
+
+    getPaymentToken(cardExpirationMonth, cardExpirationYear) {
```
davidjoy, line 56: Okay, because you want to call this function from your cybersource/service.js file, I'd probably rearrange the code to get make it possible to co-locate this function with that code.  The part where you have to save a reference to this in service.js by exporting the setter is a bit odd.

For instance, I'd favor stashing a reference to `microform` _globally_, moving this function to service.js, and hitting up that global reference for `createToken` below, rather than needing to jump through these hoops. That wouldn't be great either, but it might be clearer what's going on.

Another option - depending on how long-lived the token created by `createToken` below is - would be to have this component generate the token on component mount/microform initializatin and have it passed through (along with all the other data) to the `checkoutWithToken` function - that way it doesn't have to "reach out" at all/can just use its parameters.

src/payment/checkout/payment-form/CardDetails.jsx
davidjoy: Regarding states - the best practice is generally to have a state/status variable that uses string constants to represent different states.  It turns out it's a bit of an anti-pattern to use booleans for states, because when you find out you have more than two like this, you start creating more booleans, and then the number of combinations becomes far more than the number of valid states your app actually has. :)

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
mulby: I really like the idea of putting the `microform` instance in global scope, that's probably the simplest thing to do.

I suppose we could just pass around a state variable through the component tree to tell the credit card number field and the CVV code field to re-render when the global instance had been initialized.

mulby: Note that we have to wait until the user has filled in their credit card number and clicked "Place Order" prior to calling `createToken` so I don't think we can do this on mount/initialization.

src/payment/checkout/payment-form/CardDetails.jsx
mulby: heh, yup, that's dead code from an earlier version

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
mulby: ou - prevProps is the way to go!

src/payment/checkout/payment-form/CardDetails.jsx
mulby: Makes sense, I think we can just move them into the Render function for the form!

src/index.jsx
```diff
@@ -43,6 +55,7 @@ mergeConfig({
   APPLE_PAY_AUTHORIZE_URL: process.env.APPLE_PAY_AUTHORIZE_URL,
   APPLE_PAY_SUPPORTED_NETWORKS: process.env.APPLE_PAY_SUPPORTED_NETWORKS && process.env.APPLE_PAY_SUPPORTED_NETWORKS.split(','),
   APPLE_PAY_MERCHANT_CAPABILITIES: process.env.APPLE_PAY_MERCHANT_CAPABILITIES && process.env.APPLE_PAY_MERCHANT_CAPABILITIES.split(','),
+  WAFFLE_FLAGS: waffleFlags,
```
mulby, line 58: fedx consult: ok to store this in config?

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,87 @@
...
+    }
+    if (!this.props.captureKeyId) {
+      return;
+    }
+    if (this.timer) {
+      clearTimeout(this.timer);
+      this.timer = null;
+    }
+    if (typeof window.Flex === 'undefined') {
+      this.timer = setTimeout(this.initialize, POLLING_TIME_MS);
```
mulby, line 47: I don't feel great about this polling approach to see if the external library is loaded

src/payment/checkout/payment-form/flex-microform/FlexMicroformField.jsx
```diff
@@ -0,0 +1,85 @@
...
+                <FontAwesomeIcon icon={faQuestionCircle} />
+              </span>
+              <ReactTooltip id={`${this.props.fieldType}Help`} place="bottom" effect="solid">
+                {this.props.helpText}
+              </ReactTooltip>
+            </>
+          )}
+        <div id={`${this.props.fieldType}-container`} ref={this.element} />
+        <div className="skeleton py-3" ref={this.loadingElement} />
+        {!loading && <FontAwesomeIcon icon={faLock} className="lock-icon" />}
```
mulby, line 63: Need to display validation errors here

src/index.jsx
```diff
@@ -74,6 +87,18 @@ subscribe(APP_INIT_ERROR, (error) => {
 subscribe(APP_AUTH_INITIALIZED, () => {
   getAuthenticatedHttpClient().interceptors.response.use(responseInterceptor);
 
+  getAuthenticatedHttpClient().interceptors.request.use(async (requestConfig) => {
```
mulby, line 90: Need to enable the WAFFLE_OVERRIDE setting in ecommerce

mulby: This will send the waffle flags to the LMS as well as written

src/payment/payment-methods/cybersource/service.js
```diff
@@ -64,6 +61,84 @@ function getCardNumberDigits(cardNumber) {
...
 }
 
+function getPaymentToken(microformOptions) {
+  return new Promise((resolve, reject) => {
+    if (!window.microform) {
+      reject(Error('Microform not initialized'));
+    }
+    window.microform.createToken(microformOptions, (err, token) => {
+      if (err) {
+        reject(err);
```
mulby, line 71: This includes card number validation errors, we need to handle those explicitly and display a validation error message to the user.

https://developer.cybersource.com/api/developer-guides/dita-flex/SAFlexibleToken/FlexMicroform/api_reference/class_MicroformError.html

public/index.html
```diff
@@ -26,6 +26,7 @@
       ;NREUM.info={beacon:"bam.nr-data.net",errorBeacon:"bam.nr-data.net",licenseKey:"<%= htmlWebpackPlugin.options.newRelicLicenseKey %>",applicationID:"<%= htmlWebpackPlugin.options.newRelicApplicationID %>",sa:1}
     </script>
 
+    <script type="text/javascript" src="https://flex.cybersource.com/cybersource/assets/microform/0.11/flex-microform.min.js"></script>
```
mulby, line 29: Need to handle the case where this fails to load.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -0,0 +1,87 @@
...
+      return;
+    }
+    window.microform = new window.Flex(this.props.captureKeyId).microform({
```
mulby, line 50: This "captureKeyId" is a JWT. It is valid for 15 minutes. We need to handle the case when it expires, presumably we need to make an AJAX request to ecommerce to refresh the token and then re-initialize `window.microform`. I suspect this will wipe out whatever input the user had entered, but that's something we need to figure out.

mulby: They symptom of an expired token is that `microform.createToken()` will return an error.

src/payment/payment-methods/cybersource/service.js
```diff
@@ -64,6 +61,84 @@ function getCardNumberDigits(cardNumber) {
...
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
+    })
+    .catch((error) => {
+      const errorData = error.response ? error.response.data : null;
+      if (errorData && error.response.data.sdn_check_failure) {
+        /* istanbul ignore next */
+        if (getConfig().ENVIRONMENT !== 'test') {
+          global.location.href = `${getConfig().ECOMMERCE_BASE_URL}/payment/sdn/failure/`;
```
MatthewPiatetsky, line 111: wouldn't this leave the page before the error below can be logged?

src/index.jsx
MatthewPiatetsky: waffle_override is a different setting to allow setting any flag by including a query parameter
https://waffle.readthedocs.io/en/v0.8.1/usage.html#waffle-override
i believe as written you're using the testing property which can be used for any flag if the flag value is set to unknown and the testing checkbox is checked

public/index.html
MatthewPiatetsky: i believe we can just handle the generic case of the variable being undefined (though i'm unsure of what the product requirements are in that case, just throw and log an error?)

mulby: Yes, I think log the error and show the user the generic error message.

PR #406, opened 2020-08-19
remove unnecessary polling code for window.Flex variable and log an error if we can't initialize the flex microform

MatthewPiatetsky:


src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -25,26 +23,16 @@ class FlexMicroform extends React.Component {
...
-    if (!this.props.captureKeyId) {
-      return;
-    }
-    if (this.timer) {
-      clearTimeout(this.timer);
-      this.timer = null;
-    }
-    if (typeof window.Flex === 'undefined') {
-      this.timer = setTimeout(this.initialize, POLLING_TIME_MS);
+    if (!this.props.captureKeyId || typeof window.Flex === 'undefined') {
```
mulby, line 30: when the page first initializes it is expected that `this.props.captureKeyId` is `null` it will be non-null once the AJAX call to the backend completes.

MatthewPiatetsky: ah good point

PR #407, opened 2020-09-04
Plumb-in the capture context / public key endpoint

inventhouse:


src/payment/data/reducers.js
```diff
@@ -57,8 +60,39 @@ const basket = (state = basketInitialState, action = null) => {
...
+  loading: true,
+  loaded: false,
+  isCaptureKeyProcessing: false,
+  products: [],
+};
+
+const captureKey = (state = captureContextInitialState, action = null) => {
+  if (action !== null) {
+    switch (action.type) {
+      case fetchCaptureKey.TRIGGER: return { ...state, loading: true };
```
cpennington, line 73: Should this set `loaded: false` so that it's consistent when/if we need to reload during the timeout? Or, alternately, should there just be a single flag (`loaded`) which means `loading` when it's false, and `loaded` when it's true?

cpennington: Is `loading` used anywhere? Or are you just setting up for future work?

src/payment/data/sagas.js
```diff
@@ -34,6 +37,10 @@ function* isBasketProcessing() {
   return yield select(state => state.payment.basket.isBasketProcessing);
 }
 
+function* isCaptureKeyProcessing() {
+  return yield select(state => state.payment.isCaptureKeyProcessing);
```
mulby, line 41: Given `combineReducers` above, I'm surprised this doesn't need to be `state.payment.captureKey.isCaptureKeyProcessing`?

src/payment/data/reducers.js
```diff
@@ -57,8 +60,34 @@ const basket = (state = basketInitialState, action = null) => {
   return state;
 };
 
+const captureContextInitialState = {
+  isCaptureKeyProcessing: false,
+  flexMicroformEnabled: false,
```
mulby, line 65: I'm thinking we don't actually need to track this in the state since it's always overridden by the query string parameter in the selector

mulby (commented):
Overall looks great, just one question about the structure of the state. It's working, so I guess I'm just a bit confused about how it works.

mulby:
Also wondering aloud if now is the right time to add some unit tests around some of this logic

src/payment/data/sagas.js
inventhouse: I think you're probably right; I'll experiment some in the morning.  I believe this bug doesn't come up in normal operation, but would break the "don't start another request if one is in-flight" guard.

src/payment/data/reducers.js
inventhouse: I ended up not needing the `loading` bits and dropped them.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -72,4 +76,6 @@ FlexMicroform.defaultProps = {
   disabled: false,
 };
 
-export default FlexMicroform;
+export default connect(
+  updateCaptureKeySelector,
+)(FlexMicroform);
```
mulby, line 81: nit: It seems like this `connect` is unnecessary given that the prop is passed in explicitly here: https://github.com/edx/frontend-app-payment/pull/407/files#diff-a47751e466a3c91d8c2ab6f1f56818edL214

mulby (approved):
LGTM

PR #409, opened 2020-09-17
Cybersource token error messages

cpennington:


src/payment/payment-methods/cybersource/service.js
```diff
@@ -79,7 +79,30 @@ function getPaymentToken(microformOptions) {
 export async function checkoutWithToken(basket, { cardHolderInfo, cardDetails }) {
   const { basketId } = basket;
 
-  const paymentToken = await getPaymentToken(cardDetails);
+  const paymentToken = await getPaymentToken(cardDetails).catch((error) => {
+    if (error.reason !== 'CREATE_TOKEN_VALIDATION_FIELDS') {
```
mulby, line 83: [optional] wondering aloud if there are other types of errors that should produce user-visible feedback?

src/feedback/data/sagas.js
```diff
@@ -24,18 +24,6 @@ export function* handleErrors(e, clearExistingMessages) {
       yield put(addMessage(message.code, message.userMessage, message.data, message.messageType));
     }
   }
-  if (e.fieldErrors !== undefined) {
```
mulby, line 27: This is dead code?

mulby (approved):
LGTM

src/feedback/data/sagas.js
davidjoy: I would have expected this to still be used for field-specific validation error messages - as I recall it's how we communicate server-side validation of model fields back to the client.

davidjoy: Or are we saying we no longer want to display field errors at the top of the page?  I believe that was an a11y requirement.

src/payment/checkout/payment-form/FormInput.jsx
```diff
@@ -19,7 +22,7 @@ const FormInput = ({
       id={id}
       disabled={disabled}
     />
-    {touched && error && <span className="text-danger">{error}</span>}
+    {touched && error && <span className="text-danger">{intl.formatMessage(messages[error])}</span>}
```
davidjoy, line 25: Sanity checking - Is it the case that all error messages are coming back as message keys now, instead of formatted messages?  I see code in here elsewhere where we're definitely doing that, but I can't recall if there are other field errors that may find their way in here that were pre-formatted.

src/feedback/data/sagas.js
```diff
@@ -24,18 +24,6 @@ export function* handleErrors(e, clearExistingMessages) {
...
-    for (let i = 0; i < e.fieldErrors.length; i++) { // eslint-disable-line no-plusplus
-      const fieldError = e.fieldErrors[i];
-      yield put(addMessage(
```
davidjoy, line 30: This "addMessage" bit is what adds it to the top of the page, I believe.

src/payment/checkout/payment-form/CardHolderInformation.jsx
```diff
@@ -399,6 +399,6 @@ CardHolderInformationComponent.defaultProps = {
 };
 
 export default connect(
-  () => {},
+  null,
```
davidjoy, line 402: So this was okay but the other wasn't!?  Wacky. O_o

src/feedback/data/sagas.js
cpennington: So, when I exposed my new errors (on the CC fields) using this method, they show up as banners and on the actual fields. When I trigger the same errors on the current payment page (not using the new validation inside `payment/sagas.js`, the messages only show up by the specific field that had an error. So, either we should be putting everything in banners, or we should be putting them only on fields, but they should be consistent.

davidjoy (approved):
In general looks good.  I have some questions about the fieldErrors no longer being added as messages to the top of the page, and a sanity question around whether we're now completely normalized on passing message keys into FormInput and FormSelect, or whether there may be some stragglers, possibly in other sagas.

src/payment/checkout/payment-form/FormInput.jsx
cpennington: I believe that I got everything, but I'm not sure how I'd verify that 100%. I could add code to check whether `error` is in `messages`, and if it isn't just display `error` directly.

src/payment/payment-methods/cybersource/service.js
cpennington: https://developer.cybersource.com/api/developer-guides/dita-flex/SAFlexibleToken/FlexMicroform/api_reference/class_MicroformError.html has the errors. I didn't see any others that looked user-facing.

src/feedback/data/sagas.js
cpennington: I contest that it's unnecessary, and dumps duplicates of the error messages that are on each field onto the top of the page. But, @davidjoy suggests maybe there's an a11y issue (in which case, the other validation flow is violating that a11y consideration).

src/payment/checkout/payment-form/CardHolderInformation.jsx
cpennington: Yeah, dunno.

inventhouse (approved):

PR #410, opened 2020-09-22
suppressed pii for payment

adeelehsan:
VAN-26


waheedahmed (commented):
One question, you are adding the suppress attribute on parent div, will it hide the whole form like labels etc or just the fields?

src/payment/checkout/Checkout.jsx
```diff
@@ -111,7 +111,7 @@ class Checkout extends React.Component {
             />
           </h5>
 
-          <p className="d-flex flex-wrap">
+          <p data-hj-suppress className="d-flex flex-wrap">
```
waheedahmed, line 114: I don't think we need to hide the buttons.

waheedahmed (changes requested):
A question and some references are still unsuppressed:

1. Does add a parent div on `Field` impact on the overall design of the in any way?
2. There are still two `Field` references unsuppressed in `CardDetails.jsx` at L116 and L148.
3. There are many unsuppressed references in `CardHolderInformation.jsx` e.g. first name, last name, country etc.

src/payment/checkout/Checkout.jsx
adeelehsan: @waheedahmed Yes putting on parent would suppress its child's as well.
Regarding the unsuppressed you have mentioned, My understanding is these fields have keystroke data it would be automatically suppressed. 
I have added the suppression for the fields which have select option e.g country, year, month

waheedahmed (approved):

PR #411, opened 2020-09-22
Make capture key refresh regenerate CC / CVV microform fields

emma-green:


src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -15,20 +15,21 @@ class FlexMicroform extends React.Component {
 
     window.microform = null;
     this.state = {
-      microformStatus: DEFAULT_STATUS,
+      captureKey: { microformStatus: DEFAULT_STATUS },
```
cpennington, line 18: Don't need this change, I think?

src/payment/checkout/payment-form/flex-microform/FlexMicroformField.jsx
```diff
@@ -29,9 +30,16 @@ class FlexMicroformField extends React.Component {
   }
 
   initialize() {
-    if (!window.microform || this.field !== null || this.props.microformStatus !== STATUS_READY) {
+    if (!window.microform || this.props.microformStatus !== STATUS_READY) {
```
cpennington, line 33: I think you might be able to make this clearer by just passing `microform` as a `prop`, rather than using `window.microform`? Then you could just check if the `microform` prop went from null to non-null, and re-initialize?

cpennington: Although, maybe that would require storing the `microform` in `state`, and maybe that's not kosher? I think it's probably worth a try, though. @mulby, why did you stash `microform` on the `window` in the first place?

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -15,20 +15,21 @@ class FlexMicroform extends React.Component {
...
-    this.initialize();
+  componentDidUpdate(prevProps) {
+    this.initialize(prevProps);
   }
 
-  initialize = () => {
-    if (window.microform !== null || !this.props.captureKeyId) {
+  initialize = (prevProps = {}) => {
+    console.log("FlexMicroform Init");
+    if (!this.props.captureKeyId || prevProps.captureKeyId === this.props.captureKeyId) {
```
davidjoy, line 32: The `initialize` arguments and prop checking here may be more well-factored as:

```
componentDidMount() {
    this.initialize(this.props.captureKeyId);
}

componentDidUpdate(prevProps) {
  if (prevProps.captureKeyId !== this.props.captureKeyId) {
    this.initialize(prevProps.captureKeyId);
  }
}

initialize = (captureKeyId) => {
  // ... existing contents minus the short circuiting conditional...
}
```

That way the "Should we do this thing?" logic is contained in componentDidUpdate, where it naturally fits, and the initialize function itself just takes the one prop it actually needs (`captureKeyId`) and can just assume it's correct, and doesn't need to care if it's from the current or previous props.

src/payment/checkout/payment-form/flex-microform/FlexMicroformField.jsx
mulby: yeah, I had it in state initially, but that is an anti-pattern. I needed to stash it on window so that it could be retrieved during the saga execution when the click the submit button here:

https://github.com/edx/frontend-app-payment/blob/2c4d079907a0b077400871e59bda12875ecea888/src/payment/payment-methods/cybersource/service.js#L69

I explicitly changed it to a status prop and window var to avoid passing complex objects down the tree (per earlier feedback from FED experts).

cpennington: Ah, ok, great. Carry on, then.

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
inventhouse: This wan't the right way to do that, fixed now.

src/payment/data/sagas.js
```diff
@@ -19,8 +22,10 @@ import {
   fetchCaptureKey,
 } from './actions';
 
+import { DEFAULT_STATUS } from '../checkout/payment-form/flex-microform/constants';
```
mulby, line 25: nit: I think it would be more clear to explicitly use `STATUS_LOADING` here, it's not abundantly clear what the default is.

mulby (approved):
LGTM

cpennington (approved):

src/payment/data/sagas.js
inventhouse: Good call, changed.

PR #416, opened 2020-09-25
fix cc icon

mulby:
onChange is a real prop used by `Field` use a different name to ensure that the prop gets propagated to FlexMicroformField


cpennington (approved):

inventhouse (approved):

mulby:
I've figured out a better approach for this

PR #419, opened 2020-09-25
Make timeout messages localizable and fancy

inventhouse:


src/payment/PaymentPage.jsx
```diff
@@ -140,6 +146,12 @@ class PaymentPage extends React.Component {
             'transaction-declined-message': (
               <TransactionDeclined />
             ),
+            'capture-key-2mins-message': (
+              <CaptureKeyTimeoutTwoMinutes />
+            ),
+            'capture-key-1min-message': (
+              <CaptureKeyTimeoutOneMinute />
+            ),
```
julianajlk, line 154: Is `CaptureKeyTimeoutExplanation` not added here in the Message Codes because it's not being called in the sagas? It tags along with both 2mins and 1min message right?

src/payment/AlertCodeMessages.jsx
```diff
@@ -74,3 +74,37 @@ export const TransactionDeclined = () => (
     />
   </>
 );
+
+export const CaptureKeyTimeoutExplanation = () => (
+  <>
+    <FormattedMessage
+      id="payment.messages.key.timeout.body"
+      defaultMessage="For security, your credit card information will then need to be re-entered to complete your purchase."
```
julianajlk, line 82: I wonder if it's worth being very explicit with something like "if the purchase is not completed by the timer... then for security, your credit card information will..." or something with "this page will be refreshed and you'll need to re-enter..."

julianajlk (approved):

src/payment/PaymentPage.jsx
inventhouse: Correct; in fact `CKTExplanation` probably doesn't really need to be exported

src/payment/AlertCodeMessages.jsx
inventhouse: Perhaps; I'm trying to keep the messages short, though.  Might be something to test later.

PR #422, opened 2020-09-29
Fix UI bugs related to Flex Microform

mulby:
Should fix:

1. Credit card icons
2. iframed fields not being disabled when the user clicks "Place Order"
3. Lock icon missing


inventhouse (approved):

mulby:
Tacked on a few more fixes here:
4. All card details are in a loading state until the capture context is ready
5. State/Province field is limited to 20 characters since that's the limit in the Cybersource REST API for "administrative area"

inventhouse (approved):

PR #423, opened 2020-09-29
Only start capture key timeout if the microform is enabled

inventhouse:
Only fetching the capture key if the microform is enabled sometimes failed to do the fetch; moving the check to only affect the timeout and messages.

### DIFF
--- a/src/payment/data/sagas.js
+++ b/src/payment/data/sagas.js
@@ -154,7 +154,9 @@ export function* handleFetchCaptureKey() {
     yield put(microformStatus(STATUS_LOADING)); // we are refreshing the capture key
     const result = yield call(PaymentApiService.getCaptureKey);
     yield put(captureKeyDataReceived(result)); // update redux store with capture key data
-    yield put(captureKeyStartTimeout());
+    if (isWaffleFlagEnabled('payment.cybersource.flex_microform_enabled', false)) {
+      yield put(captureKeyStartTimeout()); // only start the timer if we're using the capture key
+    }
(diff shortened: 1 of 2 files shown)


src/payment/data/sagas.js
```diff
@@ -154,7 +154,9 @@ export function* handleFetchCaptureKey() {
     yield put(microformStatus(STATUS_LOADING)); // we are refreshing the capture key
     const result = yield call(PaymentApiService.getCaptureKey);
     yield put(captureKeyDataReceived(result)); // update redux store with capture key data
-    yield put(captureKeyStartTimeout());
+    if (isWaffleFlagEnabled('payment.cybersource.flex_microform_enabled', false)) {
```
cpennington, line 157: I think this means that if you get the flag value from the server, rather than the url, then the timer won't start. In the other places that `isWaffleFlagEnabled` is used, I have it fall back to the value from the redux store, I think. Maybe we should just have one function for `isFlexMicroformEnabled` that checks redux and waffle?

inventhouse: I had a hard time getting that from within the saga, but yes, if we can have one function for it, that would be ideal.

mulby (approved):

src/payment/data/sagas.js
inventhouse: Gonna go ahead and merge this so we can get the fix out to stop showing confusing messages in the old flow, we can look at unifying the enabled check later

PR #433, opened 2020-10-13
PCI-119: disable old payment form

inventhouse:


mulby (approved):
Looks reasonable to me!

cpennington (approved):
Can we just delete all reference to `flexMicroformEnabled`? If we need them back, we can always revert.

inventhouse:
> Can we just delete all reference to `flexMicroformEnabled`? If we need them back, we can always revert.

That's the "delete dead code" I plan to do shortly, I left them commented-out to make those places easier to find, especially where they've been turned into `if (true &&...`

PR #439, opened 2020-10-29
REV-1592 Enable Diners Club and JCB card types

inventhouse:
Note: a local test with the Diners test card did not work, I've reached out to ask if it is expected to work in the test environment
UPDATE: apparently those card types are not enabled in our testing profile, looking into updating our profiles now
UPDATE 2: I did not see how to update our test profile, but a test with the REST API succeeded without changing anything; since we are moving off SOP hopefully it will work in prod, but I updated our production SOP profile just in case.

### DIFF
--- a/src/payment/checkout/payment-form/PaymentForm.old.test.jsx
+++ b/src/payment/checkout/payment-form/PaymentForm.old.test.jsx
@@ -220,7 +220,7 @@ describe('<PaymentForm />', () => {
         ['', '', '', '', {}],
         ['', '', `${currentMonth - 1}`, `${currentYear}`, { cardExpirationMonth: 'payment.form.errors.card.expired' }],
         ['41111', '', '', '', { cardNumber: 'payment.form.errors.invalid.card.number' }],
-        ['30569309025904', '', '', '', { cardNumber: 'payment.form.errors.unsupported.card' }],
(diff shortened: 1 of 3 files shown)


mulby (approved):

dianekaplan:
This looks good!  One thought: credit-card.test.js has just one test for a supported card type (visa)- would it make sense to add a test for each icon? Or are they logically the same/unchanging so it's overkill?

inventhouse:
> This looks good! One thought: credit-card.test.js has just one test for a supported card type (visa)- would it make sense to add a test for each icon? Or are they logically the same/unchanging so it's overkill?

The codepath is pretty much the same and we change the card type logic so rarely I think just a basic test is sufficient; thanks for raising the question!

jhan217 (approved):

jhan217 (commented):
Just out of curiosity, do we know why these two card types were originally left out? Oversight or some sort of contract? pinging @sethmccann in case he has details

sethmccann:
> Just out of curiosity, do we know why these two card types were originally left out? Oversight or some sort of contract? pinging @sethmccann in case he has details

There's no business context on this decision, it appears to be an oversight in how the Discover network works and the compliance expectations of accepting their network.

PR #440, opened 2020-11-02
Re-fetch the capture key if it's already expired

cpennington:


src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -38,16 +38,25 @@ class FlexMicroform extends React.Component {
...
         },
-      },
-    });
-    this.props.dispatch(microformStatus(STATUS_READY));
+      });
+      this.props.dispatch(microformStatus(STATUS_READY));
+    } catch (err) {
+      if (err.reason === 'CAPTURE_CONTEXT_EXPIRED') {
+        this.props.dispatch(captureKeyProcessing(false));
+        this.props.dispatch(fetchCaptureKey());
```
inventhouse, line 55: We probably want to log the issue to segment so we can track this (look for `sendTrackEvent`), we also maybe need to do something so if we somehow keep using a bad key we don't just get stuck in a loop

inventhouse (dismissed):

src/payment/checkout/payment-form/flex-microform/FlexMicroform.jsx
```diff
@@ -38,16 +39,32 @@ class FlexMicroform extends React.Component {
...
+          },
         },
-      },
-    });
-    this.props.dispatch(microformStatus(STATUS_READY));
+      });
+      this.props.dispatch(microformStatus(STATUS_READY));
+    } catch (err) {
+      if (err.reason === 'CAPTURE_CONTEXT_EXPIRED' && this.state.captureContextRetryCount < 5) {
+        this.captureContextRetryCount++;
```
cpennington, line 55: Doesn't this need a `setState`?

mulby: I think this shouldn't be in state. Should just be an instance variable on the object.

PR #441, opened 2020-11-03
monkey patch Date.now() to prevent capture context rejection

mulby:
I haven't tested this at all locally -- don't have a working dev environment at the moment!


inventhouse (approved):
Gotta do what we gotta do, but ick...

PR #444, opened 2020-11-05
dont log declines as JS errors

mulby:


src/payment/payment-methods/cybersource/service.js
```diff
@@ -173,26 +173,24 @@ export async function checkoutWithToken(basket, { cardHolderInfo, cardDetails })
...
-          if (errorData && errorData.field_errors) {
-            // It's a field error
-            // This endpoint does not return field error data in a format we expect.  Fix it.
-            error.response.data = { // eslint-disable-line no-param-reassign
-              field_errors: normalizeFieldErrors(error.response.data.field_errors),
-            };
-          }
-          handleApiError(error);
-          throw error;
+        if (errorData && errorData.field_errors) {
```
inventhouse, line 185: Does this mean that (some?) field errors are (and have been) logged as JS errors?

inventhouse (approved):
"Don't log errors _with a redirect_ as JS errors", but yes, makes sense.

src/payment/payment-methods/cybersource/service.js
mulby: huh, I suppose it would, looks like that doesn't actually happen in practice since we don't see those errors in new relic?

I wonder if the backend actually catches any data validation issues that the front-end validations don't catch.

I'll rework this since that shouldn't log either.

inventhouse (approved):

PR #451, opened 2020-11-13
Flex microform ADR

mulby:


docs/decisions/0004-iframe-credit-card-handling.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Cybersource also provides an integration method called "Silent Order Post" (SOP) that allows merchants to put a <form> element on the basket page that directly posts to Cybersource allowing the merchant to avoid sending the sensitive data to back-end servers. This has been the method that frontend-app-payment has been using since it was initially created.
+
+Cybersource also provides an SDK called the "Flex Microform SDK" that allows merchants (like Open edX sites) to easily use the iframe integration method to embed the sensitive widgets directly into existing basket pages in a way that is transparent to the user.
+
+Decision
+--------
+
+frontend-app-payment will now use the Flex Microform SDK to handle credit card payments instead of SOP.
```
cpennington, line 21: We should also mention a) we aren't doing any more refactoring right now, and b) we didn't convert all uses of SOP, just the primary payment flow.

docs/decisions/0004-iframe-credit-card-handling.rst
```diff
@@ -0,0 +1,47 @@
...
+3. The SDK requires us to generate a "capture context" in order to initialize the fields. That capture context is only valid for 15 minutes. We need to refresh it every 15 minutes. When we refresh, we have to clear out any data entered into the credit card number and CVN fields. Additionally, the capture context must be generated using sensitive secrets. These secrets must be managed on the back-end so frontend-app-payment must make an AJAX request to fetch the capture context prior to rendering the credit card number and CVN fields.
+4. If Cybersource has a bug in their SDK javascript, it can prevent users from checking out.
+5. When the user clicks "Place Order" their credit card information is transformed into a transient payment token (a JWT) that can be used in lieu of their credit card number to authorize payments. However, the SOP API does not accept this type of token, so we had to switch over to using the Cybersource REST API to authorize the payment. This means that instead of direct POSTing the data to Cybersource via an HTML form, we instead make an AJAX request to the ecommerce back-end which, in turn, makes a synchronous HTTP request to Cybersource's servers.
```
cpennington, line 30: Removing SOP means that we have to handle automatic refunds if Cybersource denies but the CC processor accepts.

cpennington: Maybe worth mentioning as well that the backend lifecycle changes, because it's no longer waiting for a callback from Cybersource. Instead, it's making a call to cybersource and then processing the response to that call. That's different than all the other current payment providers.

mulby:
@cpennington I just addressed your comments!

inventhouse (approved):

PR #456, opened 2020-12-03
Update Paragon for rebranding 

julianajlk:
[REV-1604](https://openedx.atlassian.net/browse/REV-1604) rebrand work for Payment MFE, starting with updating Paragon from version 7.2.1 to 12.3.0 (some breaking changes updated for Button and StatefulButton). 

Actual branding styles changes will come in a separate PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "@edx/frontend-component-footer": "10.0.3",
     "@edx/frontend-component-header": "2.0.3",
     "@edx/frontend-platform": "1.1.9",
-    "@edx/paragon": "7.2.1",
+    "@edx/paragon": "12.3.0",
     "@fortawesome/fontawesome-svg-core": "1.2.25",
     "@fortawesome/free-brands-svg-icons": "5.11.2",
     "@fortawesome/free-regular-svg-icons": "5.11.2",
--- a/src/payment/checkout/payment-form/PaymentForm.jsx
+++ b/src/payment/checkout/payment-form/PaymentForm.jsx
@@ -208,7 +208,9 @@ export class PaymentFormComponent extends React.Component {
                   <StatefulButton
                     type="submit"
                     id="placeOrderButton"
-                    className="btn btn-primary btn-lg btn-block"
+                    variant="primary"
+                    size="lg"
+                    block
                     state={submitButtonState}
                     onClick={this.props.onSubmitButtonClick}
                     labels={{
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "@edx/frontend-component-footer": "10.0.3",
     "@edx/frontend-component-header": "2.0.3",
     "@edx/frontend-platform": "1.1.9",
-    "@edx/paragon": "7.2.1",
+    "@edx/paragon": "^12.3.0",
```
dianekaplan, line 42: The nearby convention seems to be pegging specific versions- anything to note about allowing later patch versions for paragon? (Probably because we expect more updates as the branding solidifies, but wanted to check how we decide)

MatthewPiatetsky (commented):
would be nice to see a screenshot of what this looks like

inventhouse (approved):
Looks ok to me, though I have little experience with Paragon and haven't been following the changes.

src/payment/checkout/payment-form/PaymentForm.jsx
```diff
@@ -208,7 +208,9 @@ export class PaymentFormComponent extends React.Component {
                   <StatefulButton
                     type="submit"
                     id="placeOrderButton"
-                    className="btn btn-primary btn-lg btn-block"
+                    variant="primary"
+                    size="lg"
+                    block
```
MatthewPiatetsky, line 213: is this a parameter?

jhan217 (approved):

src/payment/checkout/payment-form/PaymentForm.jsx
julianajlk: New way - we're not using classNames in Buttons and StatefulButtons anymore https://edx.github.io/paragon/components/statefulbutton

julianajlk:
> would be nice to see a screenshot of what this looks like

It should look exactly the same

package.json
julianajlk: Good call, will peg to specific 12.3.0

julianajlk: With checkout having a lot of visibility, I wouldn't want Paragon to be at a later version when you run `npm install`, which with the `^` will do that from what I understand, and that prefix allows more than just bug fixes. I don't think using `^` would allow major release changes (like for version 13 or something) but good to keep it at a specific version and not increase the version range it would allow. The caret `^` symbol is the default when you install a package with --save, that's how it ended up there

dianekaplan: that makes sense then (to change to 12.3.0)- thanks!

PR #470, opened 2021-01-22
Upgrade packages with security vulnerabilities 

julianajlk:
[REV-2084](https://openedx.atlassian.net/browse/REV-2084). 

Upgrading the below packages that have a security vulnerability:
- immer 
- yargs-parser (dependency)

Upgrading the below that is out of date:
- es-check

NOTE: `dev: true` and the version with and without `^` where removed/added automatically with `npm install package_name` or `npm install --save package_name`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5562,9 +5562,9 @@
       "dev": true
     },
     "caporal": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/caporal/-/caporal-1.3.0.tgz",
-      "integrity": "sha512-4bj21UXbEu5cF+1gVjhwwRqMhY3lR7CUTlBr6YX3uzftL4l/sbu8EoHfOLZWHr+HeiosW9fTWkQCS2UZMAk5lw==",
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/caporal/-/caporal-1.4.0.tgz",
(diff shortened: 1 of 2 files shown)


jhan217 (approved):

dianekaplan (approved):
Nothing jumps out at me, just a couple questions out of curiosity: 
- REV-2084 was for immer, how did you know to update yargs-parser and es-check?
- yargs-parser looks like a big jump (4-year old version 4.2.1 -> 20.2.4)- is there anything we do to check bigger jumps? Or if build tests pass then we're good?

julianajlk:
> Nothing jumps out at me, just a couple questions out of curiosity:
> 
> * REV-2084 was for immer, how did you know to update yargs-parser and es-check?
> * yargs-parser looks like a big jump (4-year old version 4.2.1 -> 20.2.4)- is there anything we do to check bigger jumps? Or if build tests pass then we're good?

1) I usually check for other updates whenever I am doing one of these tickets, npm will tell you with `npm audit`
2) yargs-parser is a dependency (@edx/frontend-build > new-relic-source-map-webpack-plugin > @newrelic/publish-sourcemap > yargs > yargs-parser) the advisory was to update the dependency to a later version where previously it was not a security issue to keep an old version, I believe if tests are passing and locally there are no issues, it will be ok. The version itself npm also told me 😁

PR #471, opened 2021-01-28
Added code walkthrough outline and links to ADR

inventhouse:


mulby:
Nice! One suggestion would be to include a backreference comment in code near the locations these links point to back to this doc so that when people are refactoring and grepping around in the code they are more likely to encounter this doc.

inventhouse:
> Nice! One suggestion would be to include a backreference comment in code near the locations these links point to back to this doc so that when people are refactoring and grepping around in the code they are more likely to encounter this doc.

I appreciate the suggestion, but given that a) we're planning to obsolete and retire frontend-app-payment fairly soon and b) the links reference a fairly large number of places, I don't think we should do that at this time.

mulby:
Fair enough

PR #474, opened 2021-02-26
Update README to recommend `npm ci`

inventhouse:
Recommend `npm ci` over `npm install` to minimize package_lock changes and match CI and prod builds


jhan217 (approved):
Might be nice to add a comment about why `ci` vs `install`

PR #477, opened 2021-03-16
Auto rebase renovate PRs and only create PRs on weekends

adzuci:
This will auto update renovate PRs and make renovate run on weekends.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,12 @@
 {
   "extends": [
-    "config:base"
-  ]
+    "config:base",
+    ":noUnscheduledUpdates",
+    ":semanticCommits"
+  ],
+  "rebaseStalePrs": true,
+  "schedule": [
+    "every weekend"
+  ],
+  "timezone": "America/New_York"
 }


renovate.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": [
-    "config:base"
-  ]
+    "config:base",
+    ":automergeLinters",
```
jhan217, line 4: Per discussion with @adzuci, it sounds like the automerge flags would automatically merge linting and testing package updates without it passing all of the CI checks. Not sure we'd want to do that, because it means that the person who merges after the automerges may be stuck fixing a pile of linting/testing issues on master (and potentially need to roll back their changes if they can't quickly fix forward, esp. if they didn't rebase after the automerge).

jhan217: Also, is there a list of linting / testing libraries this flag affects?

adzuci: I stole this from prospectus and tried to figure out what was considered a linter or tester library, but unfortunately couldn't see anything with the word lint in https://github.com/edx/frontend-app-payment/blob/master/package.json
https://docs.renovatebot.com/presets-default/#automergelinters

If we checked the second to last box here it would not auto merge PRs with failing tests:
https://github.com/edx/frontend-app-payment/settings/branch_protection_rules/5545303
image

Since the tests on Travis are flaky though I'll just remove this line for now.  I think it could be added back after moving tests to Github Actions and checking the right boxes.

adzuci: @albemarle Your PR from 14 months ago mentions that prospectus only auto merges testers and linters if tests pass, do you happen to know how I can make it do that here? 
image
https://github.com/edx/prospectus/pull/1430#issue-367589492

adzuci: Pretty sure I figured it out, it's in the Github repo settings, but since this repo isn't using Github Actions the tests don't always pass which means we haven't blocked PRs on tests which means it is riskier to automerge any PRs.  I removed the automerge bits.

inventhouse (approved):

PR #485, opened 2021-05-24
REV-2217: add state list for India

dianekaplan:
## Description
REV-2217: when a user selects India as the country on checkout, require the state/region field to be filled out in order to submit the order.

This field is automatically required when the country has states populated in countryStatesMap, so I've populated this list using: 
states from here: https://en.wikipedia.org/wiki/List_of_state_and_union_territory_capitals_in_India
abbreviations from here: https://kb.bullseyelocations.com/article/60-india-state-codes
(note: the second list is bigger because it also includes union territories, which we don't want to include

## Supporting information
https://openedx.atlassian.net/browse/REV-2217

## Testing instructions
I confirmed: 
- when India is selected, state is required
- when India is selected, the state dropdown is displayed matching the wiki list
- when a different county (not the other two that require states), state is not required
- order placed with Indian address goes to 'complete' in django admin, and state and country are included in billing address
- this takes advantage of existing functionality used for United States and Canada. I don't see any existing tests for countryStatesMap/getStates- not sure if I'm looking in the wrong place or if this doesn't pass the threshold to warrant tests?

Screen Shot 2021-05-25 at 9 38 04 AM


julianajlk (approved):
I'm surprised we use a static list of states and abbreviations. I wonder how often we should check those - I don't think anyone is keeping track? I know we've used libraries like pycountry in ecommerce prior to Payment MFE, would be curious to know if it's worth the effort to do something like this here too.

dianekaplan:
> I'm surprised we use a static list of states and abbreviations. I wonder how often we should check those - I don't think anyone is keeping track? I know we've used libraries like pycountry in ecommerce prior to Payment MFE, would be curious to know if it's worth the effort to do something like this here too.

Great question! I raised in our standup to weigh pros and cons, and it's not worth changing the way we do this now since we'll be moving to Stripe for the checkout page in the future- thanks for raising

PR #544, opened 2021-12-27
temp!: temporarily required ascii characters in name fields

inventhouse:
## Description
There is an issue with the SDN fallback check that is locking accounts if the name contains no ascii characters.

Added a form validation check that the name fields contain ascii letters as a short-term mitigation for [[REV-2491] Explosion in SDN hits for users with Chinese character names - JIRA](https://openedx.atlassian.net/browse/REV-2491)

It is OK for this to be a client-side check; if anyone circumvents it, they just get their account locked by the bug.

## Testing instructions
This can be tested by attempting to purchase a course and using only non-ascii letters in the name fields, for example `中国人`, adding any ascii letters will allow the fields to validate.

## Checklist
- [x] Consider PCI compliance impact and whether this PR changes how credit card information is handled.
- [x] Intend to [release and monitor](https://openedx.atlassian.net/wiki/spaces/RS/pages/1835106870/How+to+contribute+to+our+repositories) this PR promptly after merge.

### DIFF
--- a/src/payment/checkout/payment-form/PaymentForm.jsx
+++ b/src/payment/checkout/payment-form/PaymentForm.jsx
@@ -57,6 +57,10 @@ export class PaymentFormComponent extends React.Component {
 
     const errors = {
       ...this.validateRequiredFields(requiredFields),
+      ...this.validateAsciiNames(
+        firstName,
+        lastName,
+      ),
       ...this.validateCardDetails(
         cardExpirationMonth,
         cardExpirationYear,
@@ -132,6 +136,20 @@ export class PaymentFormComponent extends React.Component {
       && parseInt(cardExpirationYear, 10) === currentYear
     ) {
       errors.cardExpirationMonth = 'payment.form.errors.card.expired';
+    }
+
+    return errors;
+  }
+
+  validateAsciiNames(firstName, lastName) {
+    const errors = {};
+
+    if (
+      firstName
+      && lastName
+      && !/[A-Za-z]/.test(firstName + lastName)
+    ) {
+      errors.firstName = 'payment.form.errors.ascii.name';
     }
 
     return errors;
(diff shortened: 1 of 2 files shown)


src/payment/checkout/payment-form/PaymentForm.jsx
```diff
@@ -137,6 +141,20 @@ export class PaymentFormComponent extends React.Component {
     return errors;
   }
 
+  validateAsciiNames(firstName, lastName) {
+    const errors = {};
+
+    if (
+      firstName
+      && lastName
+      && !/[A-Za-z]/.test(firstName + lastName)
```
pshiu, line 150: Do we need to allow spaces, apostrophes, and/or hyphens too?

src/payment/checkout/payment-form/PaymentForm.messages.jsx
```diff
@@ -26,6 +26,11 @@ const messages = defineMessages({
     defaultMessage: 'This field is required',
     description: 'The form field feedback text for missing required field error.',
   },
+  'payment.form.errors.ascii.name': {
+    id: 'payment.form.errors.ascii.name',
+    defaultMessage: 'We apologize for the inconvenience but for the time being we require ASCII characters in the name field. We are working on addressing this and appreciate your patience.',
```
pshiu, line 31: Notes:
* Our users may not understand what ASCII characters mean in practice. It may be helpful to provide a hint on what characters are allowed [e.g. “we require names only have unaccented upper or lower case letters of the Latin alphabet (A–Z or a-z)”]
* If I’m reading the regex and code correctly, there may be characters in the ASCII character set (like @) that we do not accept. This may be confusing for users because the accepted characters in practice would be different from what the error message says we accept.

pshiu (approved):
Neat work! Please see a couple nits but LGTM.

src/payment/checkout/payment-form/PaymentForm.jsx
inventhouse: It just requires that there are _any_ ascii letters, but it's ok if there are other characters as well.

src/payment/checkout/payment-form/PaymentForm.messages.jsx
inventhouse: This is the wording I got from Seth, I'm inclined to just run with it; this is just temporary until we get back and fix the problem.

PR #638, opened 2022-09-21
feat: Initial use of enableStripePaymentProcessor flag

julianajlk:
[REV-3034](https://2u-internal.atlassian.net/browse/REV-3034).

Initial code to use the `enableStripePaymentProcessor` flag from the BFF ecommerce endpoint.

This is not hooked up yet to actually processing payments through Stripe, and after [REV-2999](https://2u-internal.atlassian.net/browse/REV-2999) is done, we'll use Stripe's "Place Order" button so the handle function on the button click would change location.

NOTE:
I saw an issue locally with using `enableStripePaymentProcessor` from basket where I needed to re-load sometimes, removed this to unblock REV-2999, will be added back in a following PR.

### DIFF
--- a/src/payment/checkout/Checkout.jsx
+++ b/src/payment/checkout/Checkout.jsx
@@ -64,6 +64,24 @@ class Checkout extends React.Component {
     );
   };
 
+  handleSubmitStripe = (formData) => {
+    console.log('[Zebra] handleSubmitStripe called');
+    this.props.submitPayment({ method: 'stripe', ...formData });
+  };
+
+  handleSubmitStripeButtonClick = () => {
+    console.log('[Zebra] stripeEnabled button click');
+    sendTrackEvent(
+      'edx.bi.ecommerce.basket.payment_selected',
+      {
+        type: 'click',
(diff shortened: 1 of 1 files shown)


johnnagro (approved):
looks good to me but lets get Wes to take a look before we merge.

src/payment/checkout/Checkout.jsx
```diff
@@ -64,6 +64,24 @@ class Checkout extends React.Component {
     );
   };
 
+  handleSubmitStripe = (formData) => {
+    console.log('[Zebra] handleSubmitStripe called');
```
wdrussell2015, line 68: Do we have a message logger we can use instead of console.log and make these statements debug messages? If theyre all just temp message for the sandbox then ignore this statement.

wdrussell2015 (approved):
Looks good to me.

src/payment/checkout/Checkout.jsx
julianajlk: These are just temporary for the sandbox work

PR #640, opened 2022-09-26
feat: Handle Stripe full form data on submit

julianajlk:
[REV-3001](https://2u-internal.atlassian.net/browse/REV-3001).

We're using the same form component that CyberSource uses (not specific to a payment provider), and for now we'll get the form data in a vanilla JS way.


johnnagro (approved):
a good checkpoint, lets merge and refresh the sandbox

wdrussell2015 (approved):
Ship It!

PR #641, opened 2022-10-03
refactor: Use redux-form to handle billing address form data

julianajlk:
[REV-3001](https://2u-internal.atlassian.net/browse/REV-3034).

On the previous version, we handled the billing address data with vanilla JS. Refactoring this to use `redux-form` which handles the form data in a cleaner way. It has a `handleSubmit` method which gives a handler that we can pass as a prop on submit. With this, we don't need `event.preventDefault`, and we get each field data from `values` instead of the `event.target`.

**This PR:**
- Refactors the original `handleSubmit` function by using `redux-form`'s own `handleSubmit` (they so happen to be named the same!).

**Note:** More refactoring will be needed on submit of the form in a more React way, coming soon.

**How to test this locally:**
1. Checkout branch `julianajlk/REV-3001/handle-billing-form-refactor`
2. Make sure you run `make dev.down.frontend-app-payment` in your `/devstack` directory
3. On `frontend-app-payment` run `npm ci` + `npm start`
4. Enroll in a course and upgrade
5. Billing address form should load, as well as the Stripe card elements


pshiu (approved):
LGTM! Thank you!

PR #645, opened 2022-10-05
refactor: Separate stripe client secret from cybersource capture context

wdrussell2015:
[REV-3041](https://2u-internal.atlassian.net/browse/REV-3041)

This PR: 

- Creates new redux elements for the stripe client secret instead of using the previous cybersource capture context


src/payment/data/sagas.js
```diff
@@ -143,6 +149,10 @@ export function* handleCaptureKeyTimeout() {
   yield put(fetchCaptureKey());
 }
 
+/**
+ * Redux Saga for getting the capture context for a cybersource payment
+ * @returns
```
julianajlk, line 155: Intentional?

src/payment/data/sagas.js
```diff
@@ -165,6 +175,26 @@ export function* handleFetchCaptureKey() {
   }
 }
 
+/**
+ * Redux saga for getting hte client secret key for a Stripe payment
```
julianajlk, line 179: Just a typo on "the"

src/payment/data/sagas.js
```diff
@@ -165,6 +175,26 @@ export function* handleFetchCaptureKey() {
...
+ */
+export function* handleFetchClientSecret() {
+  if (yield isClientSecretProcessing()) {
+    return;
+  }
+
+  try {
+    yield put(clientSecretProcessing(true));
+    // TODO: possibly add status for stripe elements loading?
```
julianajlk, line 188: +1 that would be great, following what we do for capture key as well

src/payment/data/sagas.js
```diff
@@ -165,6 +175,26 @@ export function* handleFetchCaptureKey() {
...
+    // const result = yield call(PaymentApiService.getClientSecret);
```
julianajlk: Is this comment intentional?
I believe you're missing updating the store with the client secret data

src/payment/data/sagas.js
```diff
@@ -164,8 +165,7 @@ export function* handleFetchCaptureKey() {
     yield put(microformStatus(STATUS_LOADING)); // we are refreshing the capture key
     const result = yield call(PaymentApiService.getCaptureKey);
     yield put(captureKeyDataReceived(result)); // update redux store with capture key data
-    // TODO: Not needed for Stripe, likely refactor to have Stripe's own hadleFetchCaptureKey
-    // yield put(captureKeyStartTimeout());
+    yield put(captureKeyStartTimeout());
```
julianajlk, line 168: I thought this was the PCI compliance for CyberSource security timeout for the credit card field, and we don't need this for Stripe

src/payment/data/sagas.js
```diff
@@ -186,7 +186,8 @@ export function* handleFetchClientSecret() {
   try {
     yield put(clientSecretProcessing(true));
     // TODO: possibly add status for stripe elements loading?
-    // const result = yield call(PaymentApiService.getClientSecret);
+    const result = yield call(PaymentApiService.getClientSecret);
+    yield put(clientSecretDataReceived(result));
```
julianajlk, line 190: Ah I see, I reviewed our PR my commit, disregard my previous comment on updating the store!

src/payment/data/service.js
```diff
@@ -56,6 +56,13 @@ export async function getCaptureKey() {
   return data;
 }
 
+export async function getClientSecret() {
```
julianajlk, line 59: Is there any difference between `getClientSecret()` and `getCaptureKey()`? If no, should we have 2?

julianajlk (approved):

src/payment/data/service.js
wdrussell2015: As of now no since the backend can only return either the capture key or the client secret.  created client secret separately so there was a clear distinction between stripe a cybersource since the capture key is used for things like microform status and the timeout warnings on the form.  I think its useful to have both for when we can have cybersource and stripe active at the same time but I can remove for now until necessary.

src/payment/data/sagas.js
wdrussell2015: fixed

src/payment/data/sagas.js
wdrussell2015: Yes. removed @returns since there is no return though

src/payment/data/service.js
julianajlk: Sounds good to keep it

src/payment/checkout/Checkout.jsx
```diff
@@ -169,7 +168,7 @@ class Checkout extends React.Component {
             {/* Apple Pay temporarily disabled per REV-927  - https://github.com/openedx/frontend-app-payment/pull/256 */}
           </p>
         </div>
-        {enableStripePaymentProcessor && options.clientSecret ? (
+        {!submissionDisabled && (enableStripePaymentProcessor ? (options.clientSecret && (
```
pshiu, line 171: Might be useful to use indents to make the logic here clearer. I'm not sure at a glance which code is in effect when.

Also, I seem to recall that we use `submissionDisabled` for multiple purposes (after submit button is pressed, while coupon is refreshing, etc). Might be worth checking that this isn't changing any user flows on the Cybersource side.

pshiu (approved):

src/payment/checkout/Checkout.jsx
wdrussell2015: fixed indentation, and changed 'submissionDisabled' to 'loaded', tested with both stripe and cybersource successfully.

PR #647, opened 2022-10-06
feat: Add Stripe Custom Actions Beta, handle Enterprise bulk enrollments data, and billing form skeleton

julianajlk:
[REV-3077](https://2u-internal.atlassian.net/browse/REV-3077).

Implementation of Custom Actions Beta (CAB) in Stripe. This requires a modification on our `onSubmit` of the form, using `updatePaymentIntent` from Stripe instead of `confirmPayment`. 
_Please note_: This is done in a non-react way for now but will change in the future.

Along with implementing CAB, we also need to address the Bulk Enrollment billing fields for Enterprise, this is how the billing form will look, with an organization field and checkbox -
Screen Shot 2022-10-12 at 12 55 00 PM

And I've added a loading skeleton for the billing forms -
Screen Shot 2022-10-12 at 3 13 26 PM

**This PR:**
- Updates the handleSubmit of the Stripe form to used `updatePaymentIntent` for CAB.
- Fixes the Place Order button placement
- Adds the Enterprise fields for Bulk Enrollment purchases to the metadata to be sent to Stripe.
- Adds a skeleton for the Billing and Card Details form to show while loading.
- Removes Zebra sandbox URL for STRIPE_RESPONSE_URL in `env` in preparation of stage testing

**Testing Done:**
Confirmed the metadata fields are sent to Stripe - https://dashboard.stripe.com/test/logs/req_mSHlxS2JmZb7kg
Screen Shot 2022-10-12 at 12 57 07 PM

Test purchase was done with `ecommerce` PR https://github.com/openedx/ecommerce/pull/3842
Redirect currently needs to be fixed but we were able to make a purchase with a coupon - 

Screen Shot 2022-10-14 at 11 58 41 AM


.env.development
```diff
@@ -30,4 +30,6 @@ APPLE_PAY_AUTHORIZE_URL='http://localhost:18130/payment/cybersource/apple-pay/au
 APPLE_PAY_SUPPORTED_NETWORKS='amex,discover,visa,masterCard',
 APPLE_PAY_MERCHANT_CAPABILITIES='supports3DS,supportsCredit,supportsDebit',
 STRIPE_PUBLISHABLE_KEY=pk_test_51Li2KoIadiFyUl1xvRRiJohVzLNtnWUYNelHjMkzaf59Mq01ZMdsCGKzh9qyRwCIHBVEv0aQPkrvdH3OJ6F6WjSv00hdOx2EMb
-STRIPE_RESPONSE_URL=http://localhost:18130/payment/stripe/checkout,
+STRIPE_BETA_FLAG=server_side_confirmation_beta_1
```
julianajlk, line 33: My understanding is that we don't need the `''` for Stripe keys/flags?

src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -81,12 +82,37 @@ function StripePaymentForm({
...
+          setMessage(result.error.message);
+        } else {
+          setMessage('An unexpected error occurred.');
+        }
+        setIsLoading(false);
+      } else {
+        // Otherwise send paymentIntent.id to your server
+        // TODO: refactor to fetch in service.js
+        const postData = JSON.stringify({ payment_intent_id: result.paymentIntent.id });
+        await getAuthenticatedHttpClient()
```
julianajlk, line 98: I've added this and the `.post()` instead of the below commented fetch because I was getting a 403 Forbidden. But I am uncertain about this

src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -81,12 +82,37 @@ function StripePaymentForm({
...
+          .post(`${process.env.STRIPE_RESPONSE_URL}`, { postData })
+          .catch(error => {
+            setMessage(error.message);
```
julianajlk, line 101: Right now the message is shown to the user for debugging purposes, but we don't want to show this IRL, since the message can be something like -
`Axios Error (Response): 400 http://localhost:18130/payment/stripe/checkout {\"field_errors\":{\"first_name\":\"This ...`

src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -81,12 +83,46 @@ function StripePaymentForm({
...
+        } else {
+          setMessage('An unexpected error occurred.');
+        }
+        setIsLoading(false);
+      } else {
+        // Otherwise send paymentIntent.id to your server
+        // TODO: refactor to fetch in service.js
+        const postData = formurlencoded({
+          payment_intent_id: result.paymentIntent.id,
+          first_name: firstName,
```
julianajlk, line 100: Checking if we need these params here because `ecommerce` is retrieving this from Stripe anyways

src/payment/checkout/Checkout.jsx
```diff
@@ -139,6 +143,10 @@ class Checkout extends React.Component {
     }
 
     const basketClassName = 'basket-section';
+
+    const loadingStripe = loading || !options.clientSecret;
+    const loadingCyberSource = loading && !enableStripePaymentProcessor;
```
colinbrash, line 148: When `loadingCyberSource` is true, `loadingStripe` will always necessarily be true as well.  Seems like something is not quite accurate here?

It also looks like you set these consts, but only use them or'd together (line 179).  If what you care about on 179 is just `(loading && !enableStripePaymentProcessor) || !options.clientSecret`, you could just set a single const to that, and name it more appropriately (I'm not sure what it's doing, is it displaying a loading state?  if so e.g., `shouldDisplayLoadingState = ...`).

src/payment/checkout/payment-form/StripePaymentForm.jsx
colinbrash: Is there a `DEBUG` flag or something you can test against to make sure this never shows up in production?  Otherwise committing something like this is a bit risky, we might forget it is there by the time we need to deploy.

src/payment/checkout/Checkout.jsx
```diff
@@ -168,6 +176,37 @@ class Checkout extends React.Component {
...
+                  <div className="skeleton py-3 mb-3" />
+                  <div className="skeleton py-3 mb-3" />
+                </div>
+                <div className="col-lg-6">
+                  <div className="skeleton py-3 mb-3" />
+                  <div className="skeleton py-3 mb-3" />
+                </div>
+              </div>
+            </>
+          ) : null}
```
wdrussell2015, line 209: Could the below stripe and cybersource elements be moved to else of the ternary here instead of null since loading is already being checked as part of the conditional.

julianajlk: I thought about that but wanted to avoid a nested ternary. But will try, maybe it will remove the issue that I have right now, when the loading skeleton is done but the Stripe fields don't load for a second (client secret is not null but the card fields take a while to load)

src/payment/checkout/payment-form/StripePaymentForm.jsx
julianajlk: I will console.log this instead

src/payment/checkout/Checkout.jsx
julianajlk: That's a good point! I suppose the JS `||` will also never reach the right part of the statement if `loading` is truthy. 
We don't have a loading state of the billing and credit card fields for CyberSource, but for Stripe, at least locally I am finding that it takes a while to get the client secret ID (required for loading the credit card fields), so I added this in, and thought I should have it for CyberSource as well, hence the 2 separate variables. I will go ahead with your suggestion, thanks Colin! 

One issue is that the page could be done loading, have the client secret, but the Stripe fields are still not loaded.

julianajlk: These are the states that currently affect this - 
CyberSource and Stripe: `loading`
Stripe: `enableStripePaymentProcessor`, `options.clientSecret`

But not having any or both of the stripe props above does not mean it's not a Stripe payment necessarily, but it's fetching this data. So `!enableStripePaymentProcessor` and `!options.clientSecret` does not mean it should be a CyberSource payment or it could mean it should!

Also, for `loading && !enableStripePaymentProcessor) || !options.clientSecret` if `enableStripePaymentProcessor` is true, and for some reason there no clientSecret (but loading is done), it would show the loading form instead of some error on obtaining the client secret.

julianajlk: One important distinction to make is that `loading` is fetching the basket data, and is not related to the `clientSecretId` fetching. One TODO is to rename to `loadingBasket` to be more clear.

src/payment/checkout/Checkout.jsx
julianajlk: Update: decided to continue not using a nested ternary (hard to read and also not recommended by the linter). I've instead created a method to return the loading skeleton to make it cleaner.

src/payment/checkout/Checkout.jsx
```diff
@@ -176,67 +213,34 @@ class Checkout extends React.Component {
...
-                disabled={submitting}
-                loading={loading}
-                loaded={loaded}
-                isProcessing={cybersourceIsSubmitting}
-                isBulkOrder={isBulkOrder}
-                isQuantityUpdating={isQuantityUpdating}
-              />
-            )
-          )}
+
```
julianajlk, line 216: Decided to separate rendering Stripe or CyberSource ternary previously done so that it's also easier to remove the CyberSource part in the future.

I am also not adding a billing form skeleton on the CyberSource payment flow since we don't have one right now, focusing on improving Zebra payment flow and not what's already there. This simplifies the conditionals I had before.

wdrussell2015 (approved):
Ship It!

colinbrash (approved):
LGTM!

src/payment/checkout/Checkout.jsx
julianajlk: Update: added `loading` so that it does show on the CyberSource payment flow. I'm curious to see how fast loading is for CyberSource that it might not even show the skeleton.

PR #650, opened 2022-10-14
fix: follow redirect sent by ecommerce on successful stripe payment

pshiu:
## Description

This PR fixes a bug where the user is not redirected to the receipt page after a successful purchase.

This is a bit of a hack because the receipt page is hit twice. Tried using Axios' maxRedirects = 0, but that still did not catch the redirect.

## Additional information

* Jira: [REV-3100](https://2u-internal.atlassian.net/browse/REV-3100)

### DIFF
--- a/src/payment/checkout/payment-form/StripePaymentForm.jsx
+++ b/src/payment/checkout/payment-form/StripePaymentForm.jsx
@@ -103,6 +103,14 @@ function StripePaymentForm({
               headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
             },
           )
+          .then(response => {
+            // TODO: This hits the receipt page twice, but Axios doesn't seem
+            // to support maxRedirects outside of NodeJS. Need to shift to a
+            // way that does not make two calls.
+            if (response.headers['content-type'].startsWith('text/html')) {
+              global.location.href = response.request.responseURL;
+            }
+          })
           .catch(error => {
             console.log('[Project Zebra] POST error: ', error.message);
           });


src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -103,6 +103,14 @@ function StripePaymentForm({
               headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
             },
           )
+          .then(response => {
+            // TODO: This hits the receipt page twice, but Axios doesn't seem
+            // to support maxRedirects outside of NodeJS. Need to shift to a
+            // way that does not make two calls.
+            if (response.headers['content-type'].startsWith('text/html')) {
+              global.location.href = response.request.responseURL;
```
wdrussell2015, line 111: On the ecommerce side could you change the response to just return the receipt page url as a string instead of using redirect() to prevent the two calls to the receipt page?

wdrussell2015 (approved):

src/payment/checkout/payment-form/StripePaymentForm.jsx
pshiu: Good point! I was trying to avoid this because of it involves a bit of refactoring given that the backend can alternatively send the user to the SDN, error, or receipt pages. I'll file a ticket for RC2 for us to imitate what Cybersource does in:
* `redirect_to_payment_error` and `redirect_to_receipt_page` in ecommerce/extensions/payment/views/cybersource.py on the backend
* src/payment/data/handleRequestError.js and `setLocation` in src/payment/payment-methods/cybersource/service.js on the frontend

grmartin (approved):

src/payment/checkout/payment-form/StripePaymentForm.jsx
pshiu: Created: [REV-3102](https://2u-internal.atlassian.net/browse/REV-3102)

PR #652, opened 2022-10-17
fix: Remove hard coded Stripe key and update env files

julianajlk:
Removing hard coded Stripe keys from `.env` files, leaving the Stripe version and beta flag value which are not specific to edX.


pshiu (approved):

grmartin (approved):
ship it 🖖

PR #653, opened 2022-10-17
feat: add initial support for stripe payment intents

pshiu:
## Description

This is a draft PR designed to solicit comments on our work to add support for Stripe Payments (or Payment Intents) to the Payment MFE.

This work was required due to a pressing business need to migrate to Stripe in a short time period.

## Additional information

* Jira: https://2u-internal.atlassian.net/browse/REV-3024
* See also: https://github.com/openedx/ecommerce/pull/3847


README.rst
```diff
@@ -215,3 +215,9 @@ If you would like to run this frontend against stage.edx.org you can run ``npm r
    :target: https://github.com/openedx/frontend-app-payment/actions/workflows/ci.yml
 .. |license| image:: https://img.shields.io/npm/l/@edx/frontend-app-payment.svg
    :target: @edx/frontend-app-payment
+
```
grmartin, line 218: still needed?

src/payment/checkout/Checkout.jsx
```diff
@@ -64,15 +72,69 @@ class Checkout extends React.Component {
...
   handleSubmitFreeCheckout = () => {
     sendTrackEvent(
       'edx.bi.ecommerce.basket.free_checkout',
       { type: 'click', category: 'checkout' },
     );
   };
 
+  renderBillingFormSkeleton() {
+    return (
+      <>
```
wittjeff, line 103: Is this a shell for tabular data?  If so it might need some ARIA roles.

src/payment/checkout/payment-form/PaymentForm.jsx
```diff
@@ -217,40 +214,12 @@ export class PaymentFormComponent extends React.Component {
...
-                        <span className="button-spinner-icon" />
-                      ),
-                    }}
-                    disabledStates={['processing', 'disabled']}
-                  />
-                )
-              }
-            </div>
-          </div>
+          <PlaceOrderButton
```
wittjeff, line 217: It looks like you're replacing a Paragon stateful button with one provided by Stripe. Is that right? I note this because conveying state info to screen reader users is a bit tricky.  No specific requests but something I'll want to test more.

pshiu:
Thank you very much for your comments! We will be adding these as tickets to our backlog. Some of our teams need this work deployed but sleeping under a flag in production, so we will merge this PR earlier than scheduled to production.

wdrussell2015 (approved):
Approving for Rollout

grmartin (approved):

src/payment/checkout/payment-form/PaymentForm.jsx
julianajlk: @wittjeff apologies for the delay in this response! We just refactored the existing button to its own component, so it can be used by different parent components. The new component still uses Paragon's `StatefulButton`

src/payment/checkout/Checkout.jsx
julianajlk: @wittjeff that's a React Fragment, it will not be rendered in the DOM.

PR #655, opened 2022-10-18
fix: false alerts and errors for stripe removed

wdrussell2015:
[REV-3057](https://2u-internal.atlassian.net/browse/REV-3057)

This pull request moves the capture context to be within the cybersource elements to prevent the cybersource timeout alerts from starting.

It also remove the useEffect on the stripe elements for setting the message since as of the custom actions beta we now use ecommerce to contact stripe instead of directly from the frontend

### DIFF
--- a/src/payment/checkout/payment-form/StripePaymentForm.jsx
+++ b/src/payment/checkout/payment-form/StripePaymentForm.jsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useState } from 'react';
 import { reduxForm } from 'redux-form';
 import formurlencoded from 'form-urlencoded';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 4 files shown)


src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -79,7 +52,7 @@ function StripePaymentForm({
       // Make sure to disable form submission until Stripe.js has loaded.
       return;
     }
-
+    setMessage('');
```
julianajlk, line 55: Do we need this here? The use of the message state has been removed on this PR, so we should also remove L139, we don't want to display just an empty string imo.

julianajlk (changes requested):

src/payment/checkout/payment-form/StripePaymentForm.jsx
julianajlk: And we wouldn't need L24 useState for this message either

julianajlk:
For the capture key, I am assuming you've tested the CyberSource flow?

src/payment/checkout/payment-form/StripePaymentForm.jsx
wdrussell2015: This is to clear any old message at the start of a submit(ie a user enters an invalid card the message will tell them their card info is incorrect then when they hit submit again the message will clear and either stay cleared or be populated by any new error message). The message state still exists and is currently updated in StripePaymentMethodHandler to show any errors from the updatePaymentIntent.  The useEffect that was removed was a watcher on the stripe state when we were confirming Payment directly to Stripe but now with the custom actions beta it doesnt make sense to have that anymore.

wdrussell2015:
> 

Yes, tested with a successful cybersource purchase

julianajlk (approved):

PR #657, opened 2022-10-19
fix: receive 200s or 400s instead of 302s from ecommerce /checkout

pshiu:
## Description
We tried having Ecommerce send 302s to redirect the Payment MFE to the receipt page. This didn't work because the library that we use for requests on the frontend, Axios, doesn't appear to support not following redirects.

This created a host of CORS and authentication issues due to logging into Ecommerce from the Payment MFE domain.

We've changed ecommerce's /payment/stripe/checkout to send a HTTP 200 or 400 instead of a 302, along with the information needed for the redirect. This commit instructs the frontend on what to do when receiving these 200 or 400s.

## Additional information

* Jira: [REV-3100](https://2u-internal.atlassian.net/browse/REV-3100)
* See also: https://github.com/openedx/ecommerce/pull/3849

## Testing instructions

Tested on local by performing a test purchase.

## Checklist
- [X] Consider PCI compliance impact and whether this PR changes how credit card information is handled.
- [X] Intend to [release and monitor](https://openedx.atlassian.net/wiki/spaces/RS/pages/1835106870/How+to+contribute+to+our+repositories) this PR promptly after merge.

### DIFF
--- a/src/payment/checkout/payment-form/StripePaymentForm.jsx
+++ b/src/payment/checkout/payment-form/StripePaymentForm.jsx
@@ -7,9 +7,14 @@ import {
   useStripe,
   useElements,
 } from '@stripe/react-stripe-js';
+
+import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { logError } from '@edx/frontend-platform/logging';
 import { AppContext } from '@edx/frontend-platform/react';
+
+import handleRequestError from '../../data/handleRequestError';
 
 import CardHolderInformation from './CardHolderInformation';
 import PlaceOrderButton from './PlaceOrderButton';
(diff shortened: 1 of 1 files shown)


src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -77,15 +100,28 @@ function StripePaymentForm({
...
-            // TODO: This hits the receipt page twice, but Axios doesn't seem
-            // to support maxRedirects outside of NodeJS. Need to shift to a
-            // way that does not make two calls.
-            if (response.headers['content-type'].startsWith('text/html')) {
-              global.location.href = response.request.responseURL;
-            }
+            setLocation(response.data.receipt_page_url);
           })
           .catch(error => {
             console.log('[Project Zebra] POST error: ', error.message);
```
julianajlk, line 106: We can remove this console.log now that you're using `setMessage`

pshiu: Done.

julianajlk (approved):
LGTM! I see that we do this for CyberSource in the `checkoutWithToken` function, for a later TODO, should we extract this out to be usable for both and have less repeated code? Or do we assume that CyberSource will be deprecated and any repeated code now will not be repeated once we remove CyberSource?

Background to this question is that there is some custom error handling for the address form that we'll need for Stripe too, and my initial idea was to move it out to an utils file so it's usable anywhere.

pshiu:
> I see that we do this for CyberSource in the `checkoutWithToken` function, for a later TODO, should we extract this out to be usable for both and have less repeated code? Or do we assume that CyberSource will be deprecated and any repeated code now will not be repeated once we remove CyberSource?
> 
> Background to this question is that there is some custom error handling for the address form that we'll need for Stripe too, and my initial idea was to move it out to an utils file so it's usable anywhere.

A utils file is a great idea. Could you add this to our backlog please if you haven't already?

PR #658, opened 2022-10-20
fix: Hide country zip code from Stripe form to avoid duplicate entry

julianajlk:
[REV-3064](https://2u-internal.atlassian.net/browse/REV-3064).

We ask the user for the country and zip code on the address form. We also ask for the same in Stripe's credit card fields.
Stripe saves the credit card country and zip code values, but we want to keep the country and zip code from the address form for SDN check, plus not ask the user for this information twice.

**Before:**
Screen Shot 2022-10-25 at 11 19 29 AM

**After (edX branding not applied):**
Screen Shot 2022-10-25 at 11 47 46 AM

**Zip Code disclaimer:**
For Stripe, zip code is not required to process payments. However, on the frontend validation, it complains if it's undefined. I added a `''` default if `undefined`. Same goes for the unit/line2 address field.
Example of a payment done with Stripe without a zip code:
https://dashboard.stripe.com/test/payments/pi_3LwYtsIadiFyUl1x0PA56G9q
It does not have "Zip check" under Payment method but that does not block payment.


src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -133,8 +133,8 @@ function StripePaymentForm({
               city,
               country,
               line1: address,
-              line2: unit,
-              postal_code: postalCode,
+              line2: unit || '',
+              postal_code: postalCode || '',
```
julianajlk, line 137: line2 and postal code are optional, but Stripe complains about it if we send `undefined`

src/payment/checkout/Checkout.jsx
```diff
@@ -72,14 +72,12 @@ class Checkout extends React.Component {
     );
   };
 
-  handleSubmitStripe = () => {
-    // TODO: We'll want to submit formData here in the next iteration
-    console.log('[Project Zebra] handleSubmitStripe called');
-    this.props.submitPayment({ method: 'stripe' });
-  };
+  // handleSubmitStripe = () => {
```
wdrussell2015, line 75: nit: is there a reason to keep this?

wdrussell2015 (approved):
looks good to me

src/payment/checkout/Checkout.jsx
julianajlk: It was added when I thought we were going to refactor _soon_ 😅 
I'll remove it, thanks Wes!

PR #663, opened 2022-10-26
fix: add ISSUE_ERROR Redux action to call <FallbackErrorMessage>

pshiu:
## Description

Currently, after pressing the checkout button, `<StripePaymentForm>` does not show any error message if the ecommerce backend throws a generic error.

This PR adds a trigger of `<FallbackErrorMessage>` for when the generic error is thrown.

The correct thing to do would have to move all the state logic in `<StripePaymentForm>` into Redux. Due to time constraints, I am implementing a hack, which is bringing Redux to `<StripePaymentForm>`.

This PR also calls `setIsLoading(false);` after the checkout button errors are handled so the user can try to re-submit if necessary.

## Supporting information

* Jira:
  * Discovered via [REV-3117](https://2u-internal.atlassian.net/browse/REV-3117) "Stage: A specific declined payment use case results in "Place order" button  stuck, page not showing any error."
  * Implements [REV-3123](https://2u-internal.atlassian.net/browse/REV-3123) "Stage: No pink error banner when attempting to buy same basket in the second tab"
* Special thanks to @grmartin for the linter fix!

## Testing instructions

* On local:
  * [X] Test purchase with decline card shows error
  * [X] Test purchase after card decline with good card is successful
* On stage:
  * [X] QA team assigned to repeat local testing on stage.
* On prod:
  * [X] QA team assigned to repeat stage testing on prod.

## Checklist
- [X] Consider PCI compliance impact and whether this PR changes how credit card information is handled.
- [X] Intend to [release and monitor](https://openedx.atlassian.net/wiki/spaces/RS/pages/1835106870/How+to+contribute+to+our+repositories) this PR promptly after merge.

### DIFF
--- a/src/payment/data/sagas.js
+++ b/src/payment/data/sagas.js
@@ -23,6 +23,7 @@ import {
   fetchCaptureKey,
   clientSecretProcessing,
   fetchClientSecret,
+  issueError,
 } from './actions';
 
 import { STATUS_LOADING } from '../checkout/payment-form/flex-microform/constants';
@@ -269,6 +270,11 @@ export function* handleSubmitPayment({ payload }) {
   }
 }
 
+export function* handleIssueError() {
+  // Show generic <FallbackErrorMessage>
+  yield call(handleErrors, {}, true);
+}
+
 export default function* saga() {
   yield takeEvery(fetchCaptureKey.TRIGGER, handleFetchCaptureKey);
   yield takeEvery(CAPTURE_KEY_START_TIMEOUT, handleCaptureKeyTimeout);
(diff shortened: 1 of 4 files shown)


src/payment/data/sagas.js
```diff
@@ -269,6 +270,11 @@ export function* handleSubmitPayment({ payload }) {
   }
 }
 
+export function* handleIssueError() {
+  // Show generic <FallbackErrorMessage>
+  yield call(handleErrors, {}, true);
```
julianajlk, line 275: This will fall under the fallback error because of the `{}` instead of passing in the actual error?

pshiu: Correct!

pshiu: Correct! This is shown here: https://github.com/openedx/frontend-app-payment/blob/c11933ca0e7224802867d6e673482db8ab0d10e9/src/feedback/data/sagas.js#L12

julianajlk (approved):
I am glad this worked!

PR #666, opened 2022-10-28
use locale to display stripe forms in another language

christopappas:
Use getLocale() to generate stripe form in a non english language. 

Screenshot after a decline:

Screen Shot 2022-10-28 at 5 04 05 PM

Screenshot after a successful purchase:

Screen Shot 2022-10-28 at 5 16 55 PM

## Checklist

### DIFF
--- a/src/payment/checkout/Checkout.jsx
+++ b/src/payment/checkout/Checkout.jsx
@@ -4,7 +4,12 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { loadStripe } from '@stripe/stripe-js';
 import { Elements } from '@stripe/react-stripe-js';
-import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import {
+  getLocale,
+  FormattedMessage,
+  injectIntl,
+  intlShape,
+} from '@edx/frontend-platform/i18n';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 
 import messages from './Checkout.messages';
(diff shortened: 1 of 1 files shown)


src/payment/checkout/Checkout.jsx
```diff
@@ -182,6 +182,15 @@ class Checkout extends React.Component {
     const shouldDisplayStripePaymentForm = !loading && enableStripePaymentProcessor && options.clientSecret;
     const shouldDisplayCyberSourcePaymentForm = !loading && !enableStripePaymentProcessor;
 
+    let stripePromise;
+    if (shouldDisplayStripePaymentForm) {
```
julianajlk, line 186: I'm curious. how this behaved locally - because `shouldDisplayStripePaymentForm` will only be true if it's not loading anymore and we have flag is true + client secret from capture-context. Does creating the Stripe instance only when this is true, does it delay Stripe's loading skeleton for the card form?

src/payment/checkout/Checkout.jsx
```diff
@@ -182,6 +182,15 @@ class Checkout extends React.Component {
...
+      stripePromise = loadStripe(process.env.STRIPE_PUBLISHABLE_KEY, {
```
pshiu, line 187: In addition to @julianajlk's point above, I think the reason we originally put `loadStripe` outside the component was to follow [this Stripe doc](https://stripe.com/docs/payments/quickstart#load-stripe):

> Make sure to call loadStripe outside of a component’s render to avoid recreating the Stripe object on every render.

One alternative would be to update the Stripe options after the i18n config is available. That looks tricky though, because per the [Stripe React docs on `options`](https://stripe.com/docs/stripe-js/react#elements-provider), it looks like `options` cannot be changed after it is set.

I suspect that the Stripe `options` would need to be strategically loaded after the i18n config is available, but before the component tree is rendered. I also suspect we load our i18n config on render, which is why this is tricky.

All this to say: this might be worth a comment as to why we are deviating from the Stripe docs. And we should keep an eye on this page's performance as we are re-loading Stripe with every render.

pshiu (approved):

src/payment/checkout/Checkout.jsx
christopappas: Added a comment in the code here. Per our discussion we can merge this and examine performance on stage. My guess is this maaay end up be an issue (thank you for identifying), but we'll see.

src/payment/checkout/Checkout.jsx
christopappas: Based on our conversation this morning, I will merge this branch and look into its performance on stage, and will revert if need be later 👍 Thank you for raising!

PR #671, opened 2022-11-02
feat: Add custom validation for Stripe payment form

julianajlk:
[REV-3148](https://2u-internal.atlassian.net/browse/REV-3148).

To keep parity with the CyberSource payment flow, we're adding the frontend/client side form validation for non-ASCII characters to the Stripe flow as well. To do this, I've refactored the validation functions we use in CyberSource payment form  to an utils file so it can be re-used in the Stripe payment form.

This also means that I've added the custom validation (instead of relying on the HTML form default validation) which should now show the error messages in red by the input that did not pass the validation.

**Before:**
Screen Shot 2022-10-27 at 5 02 20 PM

**After: styling matches existing validation styling**
Screen Shot 2022-11-01 at 2 55 48 PM

**After: Check for the countries that State and Zip Code are required -**
Screen Shot 2022-11-01 at 2 56 15 PM

**After: Check for non-ASCII characters -** 
Screen Shot 2022-11-02 at 9 47 03 AM

**_Note:_** since the first name and last name are combined in the SDN check, we have this check for both names combined. If only one is a non-ASCII character, it will pass the validation.

**This PR:**
- Refactors the validate input fields functions to utils file
- Adds the refactored functions on submit for the Stripe payment form
- Adds custom error styling on the inputs
- Update tests to account for the validation refactor
- Add the PaymentForm tests back since we now have a separate form component for the Stripe flow.

### DIFF
--- a/src/payment/checkout/payment-form/utils/form-validators.js
+++ b/src/payment/checkout/payment-form/utils/form-validators.js
@@ -1,0 +1,78 @@
+import getStates from './countryStatesMap';
+
+// eslint-disable-next-line import/prefer-default-export
+export function getRequiredFields(fieldValues, isBulkOrder) {
+  const {
+    firstName,
+    lastName,
+    address,
+    city,
+    country,
+    state,
+    cardExpirationMonth,
+    cardExpirationYear,
+    organization,
+  } = fieldValues;
+
+  const requiredFields = {
+    firstName,
+    lastName,
+    address,
+    city,
+    country,
+    cardExpirationMonth,
+    cardExpirationYear,
+  };
+
+  if (getStates(country)) {
+    requiredFields.state = state;
+  }
+
+  if (isBulkOrder) {
+    requiredFields.organization = organization;
+  }
+
+  return requiredFields;
+}
+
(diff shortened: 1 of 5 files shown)


src/payment/checkout/payment-form/utils/form-validators.js
```diff
@@ -0,0 +1,93 @@
...
+
+// eslint-disable-next-line import/prefer-default-export
+export function isPostalCodeRequired(selectedCountry) {
+  const countryListRequiredPostalCode = ['CA', 'GB', 'US'];
+  const postalCodeRequired = countryListRequiredPostalCode.includes(selectedCountry);
+
+  return postalCodeRequired;
+}
+
+export function getRequiredFields(fieldValues, isBulkOrder = false, enableStripePaymentProcessor = false) {
```
julianajlk, line 11: I added isBulkOrder and enableStripePaymentProcessor as optional, but the 2/3 places I use this function, I do need to pass in the values. The other place I use it is in the tests where I didn't want to define and pass in these args.

pshiu (approved):
Glad we got this in! Thanks for working towards a consistent UX.

PR #683, opened 2022-11-18
fix: various accessibility and stripe styling fixes

wdrussell2015:
[REV-3152](https://2u-internal.atlassian.net/browse/REV-3152)

This story addresses a few different points:
1) updated main tag to allow the skipNav link to bring the user to the main content of the page instead of having to tab through the navigation at the top.

2)added role presentation to the stripe iFrame to prevent a tab stop at the end of the cvc

3/4) Added rules to the stripe appearance field to let the inputs match the rest of the inputs on the page:
Screen Shot 2022-11-18 at 9 05 54 AM

5) Updated billing details label to express 'required'.  After speaking with Jeff Witt recommended that putting required at the header of the section instead of each field was a better solution

### DIFF
--- a/src/payment/checkout/Checkout.jsx
+++ b/src/payment/checkout/Checkout.jsx
@@ -159,6 +159,10 @@ class Checkout extends React.Component {
     const options = {
       clientSecret: this.props.clientSecretId,
       appearance: {
+        // Normally these styling values would come from Paragon,
+        // however since stripe requires styling to be passed
+        // in through the appearance object they are currently placed here.
+        // TODO: Investigate if these values can be pulled into javascript from the Paragon css files
         rules: {
           '.Input': {
             border: 'solid 1px #707070',
--- a/src/payment/checkout/Checkout.jsx
+++ b/src/payment/checkout/Checkout.jsx
@@ -178,6 +178,10 @@ class Checkout extends React.Component {
             outline: '0',
             boxShadow: '0 0 0 1px #0A3055', // $primary
           },
+          '.Label': {
+            fontSize: '1.125rem',
+            marginBottom: '0.5rem',
+          },
         },
       },
       fields: {
--- a/src/payment/checkout/Checkout.jsx
+++ b/src/payment/checkout/Checkout.jsx
@@ -156,12 +156,26 @@ class Checkout extends React.Component {
 
     // Stripe element config
     // TODO: Move these to a better home
-    const appearance = {
-      theme: 'stripe',
-    };
     const options = {
       clientSecret: this.props.clientSecretId,
-      appearance,
+      appearance: {
+        rules: {
+          '.Input': {
+            border: 'solid 1px #707070',
+            borderRadius: '0.375rem',
+          },
+          '.Input:hover': {
+            border: 'solid 1px #1f3226',
+          },
+          '.Input:focus': {
+            color: '#454545',
+            backgroundColor: '#FFFFFF',
+            borderColor: '#0A3055',
+            outline: '0',
+            boxShadow: '0 0 0 1px #0A3055',
+          },
+        },
+      },
       fields: {
         billingDetails: {
           address: 'never',
--- a/src/payment/checkout/Checkout.jsx
+++ b/src/payment/checkout/Checkout.jsx
@@ -165,7 +165,7 @@ class Checkout extends React.Component {
         // TODO: Investigate if these values can be pulled into javascript from the Paragon css files
         rules: {
           '.Input': {
-            border: 'solid 1px #707070',
+            border: 'solid 1px #707070', // $gray-500
             borderRadius: '0.375rem',
           },
           '.Input:hover': {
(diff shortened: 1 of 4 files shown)


src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -210,14 +210,15 @@ function StripePaymentForm({
       <h5 aria-level="2">
         <FormattedMessage
           id="payment.card.details.billing.information.heading"
-          defaultMessage="Billing Information"
```
christopappas, line 213: [question] does this have any effect on translations at all?

wdrussell2015: Good point, is that something that can be done pre merge or is it a post merge chore?

christopappas: I think this can be done after merge. I think it's like:
1. this PR merges to master
2. we run the push transifex job 
3. confirm there is a new (non-translated) string in transifex with "(Required)" waiting to be translated
4. _we wait some amount of time until strings are translated_
5. we run the pull transifex job to update our files here

so there will be some amount of time where the non-EN text boxes will be out of date, but based on how I understand our system, that's to be expected

src/payment/checkout/Checkout.jsx
```diff
@@ -156,12 +156,26 @@ class Checkout extends React.Component {
...
+          },
+          '.Input:focus': {
+            color: '#454545',
+            backgroundColor: '#FFFFFF',
+            borderColor: '#0A3055',
+            outline: '0',
+            boxShadow: '0 0 0 1px #0A3055',
+          },
+        },
+      },
```
julianajlk, line 178: Usually on MFE's styling is added either as utility classes from Bootstrap or through a SCSS file, which is where we would add colors like `$gray-300`, which would get pulled from Paragon as opposed to hard coding the hex color. This is because in case our design system changes the hex value for a specific color, we wouldn't have to update it here - however, this is not always followed.

I think it's ok to have them here considering Stripe is very particular, just wanted to give some edX background context. For the hex colors, where did you get them? Specifically looking at `borderColor` and the `border`? I couldn't find them under Paragon.

wdrussell2015: yeah for stripe from what I can tell the styling is to be passed in through the appearance object.  Is there a way to get those variables in the js? I tried to look to see if any of the variables were exported but it just looks like components are in paragon?

The values came from the fields above for the cardHolderInformation.  Set the stripe elements to match those.

src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -210,14 +210,15 @@ function StripePaymentForm({
...
-          description="The heading for the credit card details billing information form"
+          defaultMessage="Billing Information (Required)"
+          description="The heading for the required credit card details billing information form"
         />
       </h5>
       <PaymentElement
         id="payment-element"
         options={options}
         onReady={stripeElementsOnReady}
+        role="presentation"
```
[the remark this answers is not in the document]
wdrussell2015: Forgot to remove this.  
Was the suggested fix listed in the ticket for a tab stop at the end of the form, but on further investigation this changed nothing and it seems the tab stop only appears in safari(Tested on Chrome, Firefox, Edge, Opera, Safari) and is caused by an option in safari's preferences>advanced options: by default the option to tab through each item on a page is disabled.

src/payment/checkout/Checkout.jsx
```diff
@@ -156,12 +156,34 @@ class Checkout extends React.Component {
...
-      theme: 'stripe',
-    };
     const options = {
       clientSecret: this.props.clientSecretId,
-      appearance,
+      appearance: {
+        // Normally these styling values would come from Paragon,
+        // however since stripe requires styling to be passed
+        // in through the appearance object they are currently placed here.
+        // TODO: Investigate if these values can be pulled into javascript from the Paragon css files
```
pshiu, line 165: Oh! When previously discussing this, I hadn't understood we were hardcoding [Paragon colors](https://paragon-openedx.netlify.app/foundations/colors). I thought we were hardcoding [our company's brand colors](https://github.com/edx/brand-edx.org/blob/master/paragon/_variables.scss). In this case, they happen to be the same, so I'm not as concerned about hardcoding them.

I looked into [`getComputedStyle()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle), but it looks complicated to use in React pre-render.

@wdrussell2015, could you create a ticket for this `TODO` line with any thoughts you have for future work?

pshiu (approved):

src/payment/checkout/Checkout.jsx
wdrussell2015: https://2u-internal.atlassian.net/browse/REV-3212

PR #685, opened 2022-11-23
feat: pass product SKUs to the backend when making a purchase

christopappas:
The intent of this PR is to hand product SKUs, that were present at load time of the payment page, over to the backend to verify the basket hasn't changed in the background (to prevent a user from checking out with the wrong item).

Initial approach taken was to simply hand the products down the component tree to the place where the actual stripe calls/purchase is being made. This is the first pass to see things work to get buy-in on general design, but another thought was to add the product SKUs to state, so that instead of drilling down through the components, the payment form can snag the SKUs right from there (state).

This still needs tests in any case.

Companion ecommerce PR: https://github.com/openedx/ecommerce/pull/3876

## Checklist

### DIFF
--- a/src/payment/PaymentPage.jsx
+++ b/src/payment/PaymentPage.jsx
@@ -105,9 +105,7 @@ class PaymentPage extends React.Component {
           />
         </div>
         <div className="col-md-7 pl-md-5">
-          <Checkout
-            products={this.props.products}
-          />
+          <Checkout />
         </div>
       </div>
     );
@@ -166,7 +164,6 @@ PaymentPage.propTypes = {
   fetchBasket: PropTypes.func.isRequired,
   summaryQuantity: PropTypes.number,
   summarySubtotal: PropTypes.number,
-  products: PropTypes.array, // eslint-disable-line react/forbid-prop-types,
 };
 
 PaymentPage.defaultProps = {
@@ -174,7 +171,6 @@ PaymentPage.defaultProps = {
   isRedirect: false,
   summaryQuantity: undefined,
   summarySubtotal: undefined,
-  products: [],
 };
 
 const mapStateToProps = (state) => ({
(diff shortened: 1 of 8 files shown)


src/payment/PaymentPage.jsx
```diff
@@ -105,7 +105,9 @@ class PaymentPage extends React.Component {
           />
         </div>
         <div className="col-md-7 pl-md-5">
-          <Checkout />
+          <Checkout
+            products={this.props.products}
```
julianajlk, line 109: I believe you can get `products` from props in the Checkout component, so no need to pass this in here

src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -128,7 +129,13 @@ function StripePaymentForm({
       } else {
         // Otherwise send paymentIntent.id to your server
         // TODO: refactor to fetch in service.js
-        const postData = formurlencoded({ payment_intent_id: result.paymentIntent.id });
+
+        // generate comma separated list of product SKUs
+        const skus = products.map(({ sku }) => sku).join(',');
```
julianajlk, line 134: What data structure is `skus` in this case? If I am reading correctly, it will be a string with all the `sku` (if bundle) separated by a comma? That seems funny to me, but I am assuming that's what the BE expects

julianajlk: I clearly missed your comment 😅, I see you mention a "list" but I think this ends up being a string. Did you meant to create an array instead?

christopappas: yeah skus would be a string that has comma separated sku ids (e.g. `AB138E,DEF918`). historically i've found strings easier to deal with when passing to an endpoint, but also i originally had a list that was being urlencoded and the key in the params that made it to ecommerce was `'skus[0]'`. the `[0]` seemed weird to me, and comma separated strings are usually okay as well, so I went with that

but yeah, i meant lower case "l" list, the concept, not the data structure ;)

src/payment/PaymentPage.jsx
christopappas: oh interesting, good call. i'll drop this then

julianajlk (approved):

PR #687, opened 2022-12-01
feat: add custom message for SKU mismatch error

christopappas:
The behavior here is:
- user opens tab 1 to buy course A
- user opens tab 2 to buy course B
- user goes to tab 1 and tries to place order (which previously, and surprisingly, would buy course B)
- user sees error saying to reload the page
- user reloads page
- payment page loads with course B's information

Screen Shot 2022-12-01 at 8 42 23 AM

## Checklist
- [ ] Consider PCI compliance impact and whether this PR changes how credit card information is handled.
- [ ] Intend to [release and monitor](https://openedx.atlassian.net/wiki/spaces/RS/pages/1835106870/How+to+contribute+to+our+repositories) this PR promptly after merge.

PR #688, opened 2022-12-02
refactor: Stripe on submit to use sagas

julianajlk:
[REV-3089](https://2u-internal.atlassian.net/browse/REV-3089).

When we implemented Stripe initially, it was done using Javascript sort of React-y but not React-Redux. 
This PR is a refactor to use the `handleSubmitPayment` Redux saga on submit of the payment form, which is what all the other payment processors already use.

**This PR:**
- Removes comment and fixes something unrelated to this ticket that I found.
- Refactors the onSubmit of the form to use the submit payment saga
- Refactors error handling to use the `handleErrors` saga
- Refactors error handling for the SKU error recently added in https://github.com/openedx/frontend-app-payment/pull/687
- Removes `issueError` and `skuError` dispatchers added before when we were not using Redux
- Modifies getting state from the Redux store in `StripePaymentForm` component instead of passing in props from parent
- Fixes the loading vs. processing state of the Place Order button
- Updates/adds Jest tests

**Note:**
Loading button was showing when the user clicked on "Place Order", but it should show the processing button instead - 
Screen Shot 2022-12-13 at 7 18 54 PM

**Testing:**
For credit card errors, I used the numbers provided by Stripe - https://stripe.com/docs/testing


src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -115,87 +109,24 @@ function StripePaymentForm({
...
+        address,
+        unit,
+        city,
+        country,
+        state,
+        postalCode,
+        organization,
+        purchasedForOrganization,
       },
+      stripe,
```
julianajlk, line 125: Passing these in for now, not sure that's the best way

src/payment/payment-methods/stripe/service.js
```diff
@@ -0,0 +1,109 @@
...
+  {
+    cardHolderInfo, stripe, elements, context, products,
+  },
+  setLocation = href => { global.location.href = href; }, // HACK: allow tests to mock setting location
+) {
+  async function stripePaymentMethodHandler(result) {
+    if (result.error) {
+      // Show error in payment form
+      if (result.error.type === 'card_error' || result.error.type === 'validation_error') {
+        // setMessage(result.error.message);
```
julianajlk, line 29: This is local state in the StripePaymentForm component which I don't have access here, console logging for now

pshiu: This will mean some client-stripe Stripe validation messaging won't show to the user, correct? If we're not immediately planning to address, could we ticket remaining work here?

src/payment/payment-methods/stripe/service.js
```diff
@@ -0,0 +1,109 @@
...
+          } else {
+            // Unknown error: log and tell user.
+            logError(error, {
+              messagePrefix: 'Stripe Submit Error',
+              paymentMethod: 'Stripe',
+              paymentErrorType: 'Submit Error',
+            });
+            handleApiError(error);
+          }
+        //   setIsLoading(false);
```
pshiu, line 76: Is this being handled elsewhere?

src/payment/payment-methods/stripe/service.js
```diff
@@ -0,0 +1,109 @@
...
+          purchased_for_organization: cardHolderInfo.purchasedForOrganization,
+        },
+      },
+    },
+  });
+  stripePaymentMethodHandler(result);
+
+//   if (basket.discountJwt) {
+//     formData.discount_jwt = basket.discountJwt;
+//   }
```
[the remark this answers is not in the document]
julianajlk: Yes, I need to check if this is needed this later

src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -241,37 +202,24 @@ function StripePaymentForm({
...
-  loading: false,
   isQuantityUpdating: false,
   isProcessing: false,
   submitErrors: {},
-  products: [],
   options: null,
 };
 
 export default reduxForm({ form: 'stripe' })(connect(
   null,
```
wdrussell2015, line 224: Is the connect still needed if all that is passed is 'null'?

pshiu: Why wouldn't [`connect()`](https://react-redux.js.org/api/connect) be needed if all that is passed is null?

pshiu (approved):
Epic work, thanks @julianajlk!

src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -112,94 +120,47 @@ function StripePaymentForm({
...
-    });
-    stripePaymentMethodHandler(result);
+      });
+      onSubmitPayment({
+        paymentIntentId: result.paymentIntent.id,
+        skus,
+      });
+    } catch (error) {
+      // Show updatePaymentIntent error by the Stripe billing form fields
+      if (error.type === 'card_error' || error.type === 'validation_error') {
```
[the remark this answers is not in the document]
julianajlk: I've been wanting to remove them but I thought we discussed to keep it? I could be wrong with my memory. This could be just remnant of the initial Stripe implementation (following the docs). There is a difference in the way the `PaymentElement` does validation and how we do it with the `CardHolderInformation`, in that Stripe displays the validation error message by the field before the submit is fired, while we display the validation error only after.

This error message is displayed after the submit, but I don't see why we need it tbh.

src/feedback/data/sagas.js
```diff
@@ -15,7 +15,13 @@ export function* handleErrors(e, clearExistingMessages) {
   if (e.errors !== undefined) {
     for (let i = 0; i < e.errors.length; i++) { // eslint-disable-line no-plusplus
       const error = e.errors[i];
-      yield put(addMessage(error.code, error.userMessage, error.data, error.messageType));
+      if (error.code === 'sku-error-message') {
+        yield put(addMessage(error.code, error.userMessage, {}, MESSAGE_TYPES.ERROR));
+      } else if (error.data === undefined && error.messageType === null) {
+        yield put(addMessage('transaction-declined-message', error.userMessage, {}, MESSAGE_TYPES.ERROR));
+      } else {
```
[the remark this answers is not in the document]
julianajlk: I was wondering if we had something like `message_utils` in the backend! Thanks for the link! It would've been helpful before this was done lol. 

I will add this to a ticket so we keep error messages consistent.

src/payment/checkout/payment-form/StripePaymentForm.jsx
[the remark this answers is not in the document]
julianajlk: Update: planning to merge this before modifying for QA in Stage, will follow up in the next PR if necessary.

PR #692, opened 2022-12-06
fix: updated stripe label styling

wdrussell2015:
[REV-3031](https://2u-internal.atlassian.net/browse/REV-3031)
This PR Modifies the styling on the stripe elements to match the rest of the elements on the page
Updated Stripe fields:
Screen Shot 2022-12-06 at 10 46 57 AM

Address fields using Paragon styling:
Screen Shot 2022-12-06 at 10 47 12 AM


julianajlk (approved):
LGTM - but we should really look into how to work this in with [brand-edx.org](https://github.com/edx/brand-edx.org). The form border radius is only 0 when the above is applied and not for openedX, which means that in devstack if the edX branding is not applied, the billing address form will have rounded corners and the stripe credit card form will have square corners since it's hard coded here.

PR #693, opened 2022-12-06
fix: add error message for purchasing duplicate baskets. rename skuError

christopappas:
If you have a 2 tabs open with the same basket, you place the order on 1 tab, then go to the _other_ tab and attempt to purchase, you will get a message in the stripe element saying "An unexpected error occurred" because the paymentIntent has already succeeded after placing the first order. The problem is that this message is not useful for the user.

Now, an error message will display instructing the user to reload their page, and once they reload, they will reach the empty basket page, which is the correct behavior given that they already purchased the basket.

## Checklist

### DIFF
--- a/src/payment/checkout/payment-form/StripePaymentForm.jsx
+++ b/src/payment/checkout/payment-form/StripePaymentForm.jsx
@@ -17,7 +17,7 @@ import { injectIntl, FormattedMessage } from '@edx/frontend-platform/i18n';
 import { logError } from '@edx/frontend-platform/logging';
 import { AppContext } from '@edx/frontend-platform/react';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
-import { issueError, skuError } from '../../data/actions';
+import { issueError, basketChangedError } from '../../data/actions';
 
 import CardHolderInformation from './CardHolderInformation';
 import PlaceOrderButton from './PlaceOrderButton';
@@ -40,7 +40,7 @@ function StripePaymentForm({
   submitErrors,
   products,
   issueError: issueErrorDispatcher,
-  skuError: skuErrorDispatcher,
+  basketChangedError: basketChangedErrorDispatcher,
 }) {
   const stripe = useStripe();
   const elements = useElements();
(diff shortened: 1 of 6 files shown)


src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -123,6 +123,8 @@ function StripePaymentForm({
         // Show error in payment form
         if (result.error.type === 'card_error' || result.error.type === 'validation_error') {
           setMessage(result.error.message);
+        } else if (result.error.type === 'invalid_request_error' && result.error.code === 'payment_intent_unexpected_state') {
```
julianajlk, line 126: This if/else block is for showing a message to the user in the Stripe form (at the bottom). Is this what we want here on the new code added? If yes, can you update the comment on L123

src/payment/checkout/payment-form/StripePaymentForm.jsx
```diff
@@ -154,7 +156,7 @@ function StripePaymentForm({
               // SDN failure: redirect to Ecommerce SDN error page.
               setLocation(`${getConfig().ECOMMERCE_BASE_URL}/payment/sdn/failure/`);
             } else if (errorData && errorData.sku_error) {
-              skuErrorDispatcher();
+              basketChangedErrorDispatcher();
```
julianajlk, line 159: The variables were changed from "sku error" related to "basket changed", are you planning to keep the `sku_error` variable in ecommerce the same?

christopappas: I am planning to leave sku_error as is in ecommerce, as a sku_error is just one type of error that indicates the situation where a basket has changed

src/payment/checkout/payment-form/StripePaymentForm.jsx
christopappas: `setMessage` will show the message in the Stripe element, whereas the saga dispatch in this if/else block will show a banner to the user. Would you like me to remove the comment?

PR #696, opened 2022-12-20
feat: tell user their basket has changed when they try to purchase a …

christopappas:
…basket that is now closed

This PR does the following:
1. Moves updatePaymentIntent call to the service.js file, so there is access to the sagas (so we can dispatch the event to get the red banner with the error message displayed)
2. Removes the block of code with the error text that appears just below the Stripe elements
3. Renames SkuError to BasketChangedError, because in the SkuError case and this case, we want the same thing to happen: ask the user to refresh to see what their basket really contains (even if that's nothing)

## Checklist

### DIFF
--- a/src/payment/payment-methods/stripe/service.js
+++ b/src/payment/payment-methods/stripe/service.js
@@ -17,12 +17,55 @@ ensureConfig(['ECOMMERCE_BASE_URL', 'STRIPE_RESPONSE_URL'], 'Stripe API service'
  */
 export default async function checkout(
   basket,
-  { paymentIntentId, skus },
+  {
+    skus, elements, stripe, context, values,
+  },
   setLocation = href => { global.location.href = href; }, // HACK: allow tests to mock setting location
 ) {
+  const {
+    firstName,
+    lastName,
+    address,
+    unit,
+    city,
+    country,
+    state,
+    postalCode,
+    organization,
+    purchasedForOrganization,
+  } = values;
+
+  const result = await stripe.updatePaymentIntent({
+    elements,
+    params: {
+      payment_method_data: {
+        billing_details: {
+          address: {
+            city,
+            country,
+            line1: address,
+            line2: unit || '',
+            postal_code: postalCode || '',
+            state: state || '',
+          },
+          email: context.authenticatedUser.email,
+          name: `${firstName} ${lastName}`,
+        },
+        metadata: {
+          organization,
+          purchased_for_organization: purchasedForOrganization,
+        },
+      },
+    },
+  });
+
+  if (result.error?.code === 'payment_intent_unexpected_state' && result.error?.type === 'invalid_request_error') {
+    handleApiError(result.error);
+  }
+
   const { basketId } = basket;
   const postData = formurlencoded({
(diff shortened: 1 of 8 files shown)


src/i18n/messages/ar.json
```diff
@@ -7,7 +7,7 @@
   "payment.messages.enrollment.code.product.info.quantity.updated.header": "لقد حدّثنا كميتك.",
   "payment.messages.enrollment.code.product.info.quantity.updated.body": "في سلتك {quantity, plural,\n   one {رمز تسجيل واحد}\n   two {رمزان للتسجيل}\n   few {# رموز تسجيل}\n   many {# رمز تسجيل}\n   other {# رمز تسجيل}\n} تقدّر تكلفتها بـ {price}, ستصلك عبر البريد الإلكتروني.",
   "payment.messages.transaction.declined.body": "م نستطع معالجة دفعتك، رجاءً تحقق من معلومات الدفع الخاص بك أو تقرب من مصرفك أو مؤسستك المالية من أجل المساعدة.",
-  "payment.messages.transaction.error.sku": "Your cart has changed since navigating to this page. Please reload the page and verify the product you are purchasing.",
+  "payment.messages.transaction.error.basket_changed": "Your cart has changed since navigating to this page. Please reload the page and verify the product you are purchasing.",
```
julianajlk, line 10: I usually avoid changing the language `.json` files as they will be modified/triggered by the Transifex jobs, plus avoiding any Transifex mismatch errors from happening, just FYI. Since the ID has changed, they would have to be re-translated, and I am not sure how changing these files manually what would happen

src/payment/payment-methods/stripe/service.js
```diff
@@ -17,12 +17,53 @@ ensureConfig(['ECOMMERCE_BASE_URL', 'STRIPE_RESPONSE_URL'], 'Stripe API service'
...
+          purchased_for_organization: purchasedForOrganization,
+        },
+      },
+    },
+  });
+
+  if (result.error?.code === 'payment_intent_unexpected_state' && result.error?.type === 'invalid_request_error') {
+    handleApiError(result.error);
+  }
+
```
julianajlk, line 63: Since the error handling was removed from calling Stripe's updatePaymentIntent for `card_error` `validation_error` and `else` other than the ones you've added on L60, we don't need the `setMessage` local state anymore, could you remove it from `StripePaymentForm`? Also assuming that the cases that were removed were tested manually?

julianajlk (changes requested):
Thanks for the patience after my refactor! Let me know if you have questions on my comments

src/i18n/messages/ar.json
christopappas: ugh okay. i'll revert these changes then

julianajlk (approved):

src/payment/payment-methods/stripe/service.js
christopappas: confirming here for posterity that card error and validation error will no longer show text underneath the stripe form when Place Order is pressed. they will have the red text on the field itself, and additionally, a red banner with a generic error will be displayed (in addition to the red text) if a user tries to place an order with a card/validation error

PR #703, opened 2023-01-25
test: Added Stripe payment form tests 

wdrussell2015:
[REV-3039](https://2u-internal.atlassian.net/browse/REV-3039)
This PR uncomments previously removed tests from when we were first implementing stripe.  It also adds a new set of tests under StripePaymentForm for testing similar flows but with loading the stripe elements instead of the cybersource flex microform.


src/payment/checkout/Checkout.test.jsx
```diff
@@ -106,19 +109,24 @@ describe('<Checkout />', () => {
...
-    //     category: 'checkout',
-    //     paymentMethod: 'Credit Card',
-    //     checkoutType: 'client_side',
-    //     flexMicroformEnabled: true,
-    //   });
-    // });
+    it('submits and tracks the payment form', () => {
+      expect(sendTrackEvent).toHaveBeenCalledWith('edx.bi.ecommerce.payment_mfe.payment_form_rendered', {
+        ...getPerformanceProperties(),
+        paymentProcessor: 'Cybersource',
```
julianajlk, line 115: Is see that the payment processor is Cybersource, I am surprised this is passing considering there is nothing about the flag to determine if it should render which payment form?

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
```diff
@@ -0,0 +1,283 @@
...
+
+  describe('load stripe', () => {
+    it('creates the payment element', () => {
+      expect(mockElements.create).toHaveBeenCalledWith('payment', null);
+      expect(paymentElement.length > 0).toBe(true);
+    });
+  });
+
+  describe('getRequiredFields', () => {
+    it('returns expected required fields', () => {
```
julianajlk, line 122: I see that these tests are like the [PaymentForm.test.jsx tests](https://github.com/openedx/frontend-app-payment/blob/master/src/payment/checkout/payment-form/PaymentForm.test.jsx#L78). This `getRequiredFields` test suite is just checking if the fields are being rendered but not necessarily checking for the "required" label or anything like that?

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
```diff
@@ -0,0 +1,283 @@
...
+      const testFormValues = [
+        {
+          firstName: '',
+          lastName: '',
+          address: '',
+          city: '',
+          country: 'UK',
+          cardExpirationMonth: '',
```
julianajlk, line 130: We don't have these fields as part of our custom form for Stripe, should they be removed? CyberSource does have the custom credit card fields.

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
```diff
@@ -0,0 +1,283 @@
...
+        </IntlProvider>
+      ));
+
+      const formValues = {
+        firstName: '',
+        lastName: '',
+        address: '',
+        city: '',
+        country: 'UK',
+        cardNumber: '',
```
julianajlk, line 193: Same comment as above, could we remove the credit card related fields since we don't get the values for them (goes directly to Stripe)?

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
```diff
@@ -0,0 +1,283 @@
...
+          expect(submitStripePayment).not.toHaveBeenCalled();
+        } else {
+          expect(() => submitButton.simulate('click')).not.toThrow();
+        }
+      });
+    });
+  });
+
+  describe('validateRequiredFields', () => {
+    it('returns errors if values are empty', () => {
```
julianajlk, line 272: Is it worth testing for Stripe if the required fields for the certain countries is being required onSubmit?

julianajlk (commented):
Thanks for uncommenting the other tests - we forgot to add them back!
Let me know if you have questions to my comments.

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
```diff
@@ -0,0 +1,283 @@
...
+      ];
+
+      testFormValues.forEach((formValues) => {
+        const requiredFields = formValidators.getRequiredFields(formValues);
+        if (formValues.country) {
+          const { optionalField, ...expectedRequiredFields } = formValues;
+          expect(requiredFields).toEqual(expectedRequiredFields);
+        } else {
+          const { optionalField, ...expectedRequiredFields } = formValues;
+          expect(requiredFields).toEqual(expectedRequiredFields);
```
pshiu, line 163: This `if...else` appears to execute the same statements for every condition; is this correct? What motivated the addition of these conditions?

wdrussell2015: Just an oversight I forgot to combine. fixed

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
wdrussell2015: like state and zip? good call Ive added those into the getRequiredFields

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
wdrussell2015: removed

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
wdrussell2015: removed

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
wdrussell2015: Its not checking the elements at all, this is checking that when we hit submit we have included all the fields weve defined as required.  Updated this to include postal and state based on your other comment below.

src/payment/checkout/Checkout.test.jsx
wdrussell2015: the cybersource payment form is shown by default if there is no flag.  All the cybersource elements check for is not loading and not stripe enabled Checkout.jsx lines 220 and 280.

julianajlk: Gotcha, that's true, when the flag is `false` and not loading it will render the CyberSource, great thanks!

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
julianajlk: Great, thanks! One thing to note is that in one of the example `testFormValues`, UK should have postal code since it's required in the Stripe implementation.

I see these examples were taken from the `PaymentForm` tests, but would be good to have this for Stripe with an example of a country where state and postal code is required (aka. US), where only postal code is required (UK) and where only state is required (India)?

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
julianajlk: Nice catch! It's interesting that in the `PaymentForm` test, it has the `if formValues.country` but all of examples in the test form values have a country, so my understanding is that the test will never reach the `else`.

Combining makes sense.

julianajlk (approved):

src/payment/checkout/payment-form/StripePaymentForm.test.jsx
wdrussell2015: Added

PR #708, opened 2023-02-10
Update: added sequence diagram for ecommerce and stripe

zubair-ce07:
Description:

- Added updated sequence diagram to update frontend-app-payment ADR-5

## Other information
Ticket [REV-3390](https://2u-internal.atlassian.net/browse/REV-3390)


julianajlk (commented):
Thanks for adding this! I have a couple of questions - 
1) For the parallel requests section, those are happening asynchronously/not in order in the location it's placed? I saw that before on the CyberSource flow, the call to ecommerce's capture-context was placed in this section, but for this Stripe version, its not (`getClientSecret`).
2) There is a call from ecommerce to Stripe to get the billing info (sent from the MFE in `updatePaymentIntent`), should it be added in this flow?
3) At the very bottom of the flow, `Confirm PaymentIntent ID` - should this be just `Confirm PaymentIntent`?
4) We don't include SDN and embargo checks in this flow chart? I see the CyberSource flow chart didn't either.

julianajlk (approved):

PR #746, opened 2023-05-22
Mrazadar/PON-14/checkout-page-design-edits

mrazadar:
- Checkout page design edits and copy changes
- refactor: removed the new 18n translations files
- test: fixed the failing tests
- feat: removed the fraction zero digits 
- feat: better error handling 
- fix: Button states fixed


NawfalAhmed (approved):

src/subscription/alerts/SubscriptionAlerts.jsx
```diff
@@ -15,8 +15,8 @@ export const SubscriptionAlerts = () => (
   <AlertList
     messageCodes={{
       empty_subscription: (<EmptySubscriptionMessage />),
-      'embargo-error': (<EmptySubscriptionMessage />),
-      'basket-changed-error': (<BasketChangedError />),
+      embargo_error: (<EmptySubscriptionMessage />),
```
julianajlk, line 18: Can we add `<FallbackErrorMessage />` instead until Product decides what we should display to the learner in this case?

julianajlk (approved):
LGTM! Curious on the reason to change `confirmation_status` to `status`?

src/subscription/alerts/SubscriptionAlerts.jsx
mrazadar: > LGTM! Curious on the reason to change `confirmation_status` to `status`?

https://2u-internal.atlassian.net/browse/PON-226

mrazadar: > Can we add `<FallbackErrorMessage />` instead until Product decides what we should display to the learner in this case?

done

PR #747, opened 2023-05-23
fix: Minimum stripe amount fix

mrazadar:
Fix: stripe price bug fix
Fix the broken tests
Reverted the timeout change as the axios default timeout is 0, that means unlimited waiting.


src/subscription/checkout/StripeOptions.js
```diff
@@ -10,8 +10,10 @@ export const getStripeOptions = ({ currency, price }) => ({
    * NOTE: this is important to set the right subscription price as defined within stripe subscription
    * otherwise card details will not load properly
    * https://dashboard.stripe.com/test/subscriptions/sub_1Mtd3sH4caH7G0X1dvzlMT3o
+   * Stripe consider amount in cents and minimum amount it expects is 0.50 cents
```
julianajlk, line 13: Can you change this to `* Stripe expects the amount in cents for USD currency, and minimum amount it expects is 0.50 cents`

julianajlk: We should ask QA to test this in localized currencies. BE sends the price in USD, so I think we're good here but just in case

julianajlk (approved):
Thanks for debugging this issue, I'm so glad we found it 👏

NawfalAhmed (approved):
Looks good

MuhammadMattiullah (approved):

src/subscription/checkout/StripeOptions.js
mrazadar: done.

PR #750, opened 2023-05-31
feat: handled checkout errors

mrazadar:
Handled more error alerts according to the ticket description.

Ticket: https://2u-internal.atlassian.net/browse/PON-249
Screenshot 2023-05-31 at 6 08 54 PM

### DIFF
--- a/src/subscription/alerts/ErrorMessages.jsx
+++ b/src/subscription/alerts/ErrorMessages.jsx
@@ -1,6 +1,49 @@
 // import FallbackErrorMessage from '../../feedback/FallbackErrorMessage';
+import React from 'react';
+import { getConfig } from '@edx/frontend-platform';
+import { Hyperlink } from '@edx/paragon';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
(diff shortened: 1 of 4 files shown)


src/subscription/alerts/ErrorMessages.jsx
```diff
@@ -1,6 +1,49 @@
...
+import React from 'react';
+import { getConfig } from '@edx/frontend-platform';
+import { Hyperlink } from '@edx/paragon';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+
 // eslint-disable-next-line import/prefer-default-export
-export const EmptySubscriptionMessage = () => (
-  <>
-  </>
+export const EmbargoErrorMessage = () => (
```
aht007, line 8: Are we sure that for subscription programs we want to show embargo error this way?
Currently in ecommerce we are just showing fallback error. I might be missing something here if we have decided to do it this way now for subscriptions.

aht007 (approved):
Added a comment. Otherwise LGTM.

PR #752, opened 2023-05-31
fix: Internationalization of ErrorPage when Cookies aren't available

grmartin:
Checking to see if we have a valid Locale in English (really as a test of if messages are set) and if not, calling in a blank localization so the page doesnt simply crash.


grmartin:
This was reported yesterday as: https://github.com/openedx/frontend-template-application/issues/590

christopappas (approved):

PR #774, opened 2023-07-17
feat: Add Coordinator Waffle Flags

grmartin:
- fix: swapping out getQueryParameters with searchParams

  - `getQueryParameters` has been deprecated, so the code has been modified to use `URL(...).searchParams`.

  - Moving Waffle Flag handling to an external file.

  - 100% Coverage for Waffle Flag processing as well as interception.

    Also allowlisted a vulnerability.

    > Note: Used API 2 for Jest in its Tabular format. Nice for simpler values. and less eslint/visual formatting issues.

- fix: Merge COMMERCE_COORDINATOR_BASE_URL in Config from Env

  - Includes moving service code so CC calls are at the end of the file.

- chore: Fixing some excluded code coverage files

  - Permitting `src/data` to be covered
    - excluding the specific files without coverage.
  - Added a simple rootSagas Test

- fix: Moving isWaffleFlagEnabled with its sibling function

- feat: Accepting Waffle Flags from the Environment 

  - A simple hijack of URL parsing to allow flags from the environment. This should ease development.

- feat: Adding isCommerceCoordinatorEnabled

  - Checks Waffle Flags to see if were in CC mode.

    Additional testing and test refactoring.

- chore: Doc updates and moving Waffle Flags from magic strings to an Enum

- chore: Set CC Waffle Flag to use the MS1 Switch for Backend Selection

- feat: Swapping between Ecomm IDA and Coordinator based on Flags

  - Testing updates including returning promises from the `perform` functions so that async work doesn't time out Jest
  - Documentation Updates
  - Setting payment processing dialog to never show unless we are in CC Mode
  
  JIRA: https://2u-internal.atlassian.net/browse/THES-49


src/payment/data/sagas.test.js
```diff
@@ -41,14 +41,15 @@ import '../__factories__/basket.factory';
 
 import * as cybersourceService from '../payment-methods/cybersource';
 import {
-  DEFAULT_PAYMENT_STATE_POLLING_DELAY_SECS,
   DEFAULT_PAYMENT_STATE_POLLING_MAX_ERRORS,
   PAYMENT_STATE,
-  POLLING_PAYMENT_STATES,
+  POLLING_PAYMENT_STATES, WAFFLE_FLAGS,
```
grmartin, line 46: fix for this formatting issue is already complete and will be pushed with other preliminary feedback.

src/payment/data/utils.js
```diff
@@ -288,3 +284,14 @@ export function createCustomRoutine(name, addtlStages, keepDefaults = true) {
...
+/**
+ * Determines if the MFE should follow the Commerce Coordinator Flows
+ *
+ * This checks a waffle flag and assumes the CommerceCoordinator has set the flag if the order type is right.
+ *
+ * @returns {boolean}
+ */
+export function isCommerceCoordinatorEnabled() {
+  return isWaffleFlagEnabled(WAFFLE_FLAGS.COMMERCE_COORDINATOR_ENABLED);
+}
```
pshiu, line 297: Can we call `isWaffleFlagEnabled(WAFFLE_FLAGS.COMMERCE_COORDINATOR_ENABLED)` directly? Worried this is too much wrapping.

src/data/waffleFlags.js
```diff
@@ -0,0 +1,55 @@
+import { getConfig } from '@edx/frontend-platform';
```
pshiu, line 1: Can we add a link to [Feature Toggle HOWTO](https://github.com/openedx/frontend-app-payment/blob/master/docs/how_tos/feature_toggle.rst) somewhere appropriate in this file?

src/payment/data/constants.js
```diff
@@ -1,11 +1,48 @@
+/**
+ * Order Types
```
pshiu, line 2:
```suggestion
 * Order Types are the possible types of products a learner can purchase.
 *
 * Each product type may have different user flow requirements.
 *
 * For example, bulk enrollments allow users to change the quantity of a product.
```

src/payment/data/constants.js
```diff
@@ -1,11 +1,48 @@
...
+ * @readonly
+ * @enum {string}
+ */
 export const ORDER_TYPES = {
+  /**
+   * The Redemption of a Single Course Enrollment Code
+   *
+   * A bulk-purchases on a single course, usually B2C for 100% off of course seats.
```
pshiu, line 10: Spelling:

```suggestion
   * A bulk purchase on a single course, usually B2C for 100% off of course seats.
```

src/payment/data/constants.js
```diff
@@ -1,11 +1,48 @@
...
+   */
   BULK_ENROLLMENT: 'Enrollment Code',
+
+  /**
+   * Program Entitlement
+   *
+   * This means someone has purchased a full program containing the course, so they have already paid but may not
+   * actually be enrolled in a run.
```
pshiu, line 18:
```suggestion
   * A purchase of a program, which are multiple products called "course entitlements".
   * 
   * A course entitlement allows a learner to redeem the entitlement for a seat in a course at a future time.
```

src/payment/data/constants.js
```diff
@@ -15,6 +52,9 @@ export const CERTIFICATE_TYPES = {
  * @see POLLING_PAYMENT_STATES
  *
  * @note: This enum is unique to Commerce Coordinator backend.
+ *
+ * @readonly
+ * @enum {string|Symbol}
  */
 export const PAYMENT_STATE = (((webserviceEnum = {
```
pshiu, line 59: Nit: Can we replace `// Our Additions` on line 80 below with an explanation that the payment states listed below are added by Payment MFE to the backend's list of possible payment states?

src/payment/data/service.js
```diff
@@ -94,18 +93,49 @@ export async function getDiscountData(courseKey) {
...
-            order_uuid: basketId,
-          },
-      },
-    )
-    .catch(handleBasketApiError);
-  return data;
+// Commerce Coordinator Specific Calls
+const CC_BASE = getConfig().COMMERCE_COORDINATOR_BASE_URL;
+
+export class CommerceCoordinator {
```
pshiu, line 99: In general, we should try to match Coordinator's API to frontend-app-payment rather than vice-versa. Could you document what we need to do to get Coordinator working with frontend-app-payment?

If there is a refactor we can do to reduce the amount of multiplexing here, that would be awesome.

src/data/waffleFlags.test.js
```diff
@@ -0,0 +1,167 @@
...
+      })
+      .then(done);
+  });
+});
+
+/**
+ * Set our JSDOM Window and Document Location
+ *
+ * **NOTE**: You *MUST* include your tests async DoneCallback as the final action of your `perform`'s closure.
+ * failure to do so will lead to test timeouts.
```
pshiu, line 50: Question: Why is this comment located here, as opposed to the test files where `done()` is called?

Asking because placing this comment here makes it sound like there is something in this function that requires a done callback, but I'm not sure what it is.

My understanding was that the done callback pattern is only necessary when we have a test with a callback we don't await. But what callback are we are not awaiting in our tests - could you point me to them? Or am I missing something?

.env
```diff
@@ -38,4 +38,5 @@ STRIPE_RESPONSE_URL=null
 STRIPE_DEFERRED_INTENT_BETA_FLAG=elements_enable_deferred_intent_beta_1
 SUBSCRIPTIONS_BASE_URL=null
 ENABLE_B2C_SUBSCRIPTIONS=false
-COMMERCE_COORDINATOR_BASE_URL=null
\ No newline at end of file
+COMMERCE_COORDINATOR_BASE_URL=null
+WAFFLE_FLAGS=""
```
pshiu, line 42: We should note somewhere that waffle flags coming via query parameters override any `WAFFLE_FLAGS` set here.

src/payment/data/selectors.js
```diff
@@ -90,9 +91,11 @@ export const paymentProcessStatusSelector = state => (state[storeName].basket.pa
  * @param state
  * @return boolean
  * @see POLLING_PAYMENT_STATES
+ * @see WAFFLE_FLAGS.COMMERCE_COORDINATOR_ENABLED
  */
 export const paymentProcessStatusShouldRunSelector = state => (
   POLLING_PAYMENT_STATES.includes(paymentProcessStatusSelector(state))
+  && isWaffleFlagEnabled(WAFFLE_FLAGS.COMMERCE_COORDINATOR_ENABLED)
```
pshiu, line 98: The collision of the constant `WAFFLE_FLAGS` with the environmental variable `WAFFLE_FLAGS` set in .env & sister files got confusing for me. What do you think of renaming one of them?

src/payment/data/service.js
```diff
@@ -3,12 +3,68 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
...
+import { isWaffleFlagEnabled } from '../../data/waffleFlags';
+import { WAFFLE_FLAGS } from './constants';
 
 ensureConfig([
   'ECOMMERCE_BASE_URL',
   'LMS_BASE_URL',
 ], 'payment API service');
 
+/* eslint-disable no-use-before-define */ /* Chicken vs Egg...
+    but it is preferred the interpreter pickup errors rather than use magic strings */
```
pshiu, line 15: Might be worth elaborating what is the chicken vs. egg problem with `.name`.

src/payment/data/service.js
```diff
@@ -3,12 +3,68 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
...
+
+/**
+ * Ecommerce IDA/Coordinator IDA URLs keyed by the function name they are meant for.
+ *
+ * @type {{[p: string]: string}}
+ */
+const urlsByFunction = {
```
pshiu, line 22: It sounds like this will require filing tickets for remapping the Coordinator & API specs to use similar endpoints names as Ecommerce?

What if we want to name something differently in the future?

src/payment/data/service.js
```diff
@@ -3,12 +3,68 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
...
+  /* Shared common paths */
+  [getClientSecret.name]: '/bff/payment/v0/capture-context', // no forward slash at the end or we will explode
```
pshiu, line 24: "We will explode" is colorful, but not really accurate. Could we say that ecommerce will 404?

pshiu:
@grmartin, could you add a commitlint prefix to the PR title & add a link to the ticket in the description of the PR?

.env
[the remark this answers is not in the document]
grmartin: habit. ill fix

grmartin: agreed, no docs have been updated. with this work. There is a CC side ticket for this, ill mod it to be both.

src/data/waffleFlags.test.js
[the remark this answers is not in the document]
grmartin: Fair, but that is arguably a flaw in our package.json file and we all should be when able (makes jumping in and fixing things easier) But happy to fix.

src/payment/data/reducers.js
```diff
@@ -141,6 +135,7 @@ const clientSecretInitialState = {
   clientSecretId: '',
 };
 
+// CC & Ecomm IDA Backends, Stripe
```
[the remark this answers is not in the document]
grmartin: fair.

src/payment/data/sagas.js
```diff
@@ -41,8 +44,13 @@ import { checkout as checkoutPaypal } from '../payment-methods/paypal';
...
 import { checkout as checkoutStripe } from '../payment-methods/stripe';
 import { paymentProcessStatusShouldRunSelector } from './selectors';
-import { PAYMENT_STATE, DEFAULT_PAYMENT_STATE_POLLING_DELAY_SECS, POLLING_PAYMENT_STATES } from './constants';
+import {
+  PAYMENT_STATE,
+  DEFAULT_PAYMENT_STATE_POLLING_DELAY_SECS,
+  POLLING_PAYMENT_STATES,
+} from './constants';
 import { generateApiError } from './handleRequestError';
+import { getCurrentPaymentState } from './service';
```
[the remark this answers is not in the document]
grmartin: Yes, my IDE is configured to not use *s id assume, since its considered bad form..

src/payment/data/selectors.js
grmartin: Environment is simpler, ill do it there... `ENABLED_WAFFLE_FLAGS`?

src/payment/data/service.js
grmartin: I assumed general knowlege of our tools, ill simplify.

src/payment/data/service.js
grmartin: We did that in PR 100. The documentation is wrong now. Was your suggestion.

src/payment/data/service.js
grmartin: i dont recall the exact error, but im coonfusing whether its BAD_METHOD or FILE_NOT_FOUND. But i will clarify.

src/data/waffleFlags.test.js
```diff
@@ -0,0 +1,167 @@
...
+describe('getWaffleFlags', () => {
+  it('should default to document.location when empty', () => {
+    // We have to use the Window's origin, otherwise: SecurityError: replaceState cannot update...
+    //   This was WAY TOO HARD FOUGHT.
+    const testLocation = `${window.origin}/dox.asp?${WAFFLE_PREFIX}xyzzy=on`;
+
+    return performWithModifiedJSDOMLocation(testLocation, () => {
+      const result = processUrlWaffleFlags();
+      expect(result).toStrictEqual({ xyzzy: true });
+    });
```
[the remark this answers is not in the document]
grmartin: No, the machoism for handling history in Jest doesnt work right as it mimics a full DOM via JSDOM. Wes and i tried some 7 different forms, none took. This is why we had to work with the history scoping mechanism as well because it throws various `SecurityError: `'s, such as the one depper in the stack called `replaceState` even though we never touch it directly, its minicing the Mozilla Docs (i assume) and such.

.env
grmartin: [THES-242](https://2u-internal.atlassian.net/browse/THES-242)

src/data/waffleFlags.test.js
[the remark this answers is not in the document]
grmartin: Also theres needed cleanup, to reset the DOM for the next call, this requires the caller to know nothing of that, and it 'just works'.

src/data/waffleFlags.test.js
grmartin: I have added to the documentation, feel free to open this again if its confusing.

src/data/waffleFlags.test.js
```diff
@@ -0,0 +1,167 @@
...
+  const cleanup = () => {
+    mergeConfig({ WAFFLE_FLAGS: initialConfig });
+  };
+
+  return executePossiblyAsyncFunction(perform, cleanup);
+}
+
+describe('getWaffleFlags', () => {
+  it('should default to document.location when empty', () => {
+    // We have to use the Window's origin, otherwise: SecurityError: replaceState cannot update...
```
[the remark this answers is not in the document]
grmartin: In code update, eopen if still an issue :)

pshiu (approved):

PR #776, opened 2023-07-20
feat: Added stripe redirect content-security-policy meta tag

mrazadar:
This pull request uses `react-helmet` package to add Stripe content security policy meta tag to allow bank sites to redirect to payment MFE after successful 3DS. 

https://stripe.com/docs/security/guide#content-security-policy


NawfalAhmed (approved):
Nice work 👍

PR #778, opened 2023-07-20
revert: removed the CSP meta tag

mrazadar:
This PR reverted the change for adding a Content Security Policy meta tag in the index.html. We are going to update the `terraform` CSP headers for payment MFE.


julianajlk (approved):
Comment for future self, we're making a CSP modification via terraform instead, as this blocks other frame-src that are not Stripe.

PR #781, opened 2023-07-27
fix: Checkout Bulk purchase 'Max' alignment fixed

mobeenali12:
Checkout page has a design alignment issue under Quantity section. Just added and removed some style's classes of components.
Ticket: [REV-3615](https://2u-internal.atlassian.net/browse/REV-3615)


zubair-ce07 (approved):
LGTM

PR #788, opened 2023-08-03
feat: sending program_uuid with stripe-checkout-complete

mrazadar:
As per request by Chris/Julian sending `program_uuid` with `stripe-checkout-complete` post. 
Plus removed some unnecessary logs


julianajlk (approved):
Thank you!

PR #819, opened 2023-10-20
chore: Update to the new version of brand-openedx in the new scope.

feanil:
Part of https://github.com/openedx/axim-engineering/issues/23

This updates the `@edx/brand` alias to point to the `brand-openedx` package at
the `openedx` scope. This does not impact imports because this package is used
via an alias.


julianajlk (approved):
The caret is intentional I'm assuming, which should be ok for minor/patch changes, I see ecommerce MFE already uses it.

PR #849, opened 2024-02-07
feat: Use country-state-city to populate state dropdown in StateProvinceFormInput

julianajlk:
[REV-3842](https://2u-internal.atlassian.net/browse/REV-3842).

Removing the hard coded states for Canada, US, and India in `countryStatesMap.js` and using the `country-state-city` library to populate states for all countries that have states in the ISO 3166-2 list. Sonic version of the payment page will use this library as well. This is done in preparation for offering Dynamic Payment Methods, where the Stripe API requires state/zip code for BNPL payment methods.

**_Note:_**: If the country has states/provinces/regions/countries/districts, state field will be populated and will become a REQUIRED field.


zubair-ce07 (approved):

zubair-ce07:
LGTM 👍

PR #850, opened 2024-02-09
refactor: Add Dynamic Payment Methods handling

julianajlk:
**_Update:_** replaced with forked edx version - https://github.com/edx/frontend-app-payment/pull/4
[REV-3830](https://2u-internal.atlassian.net/browse/REV-3830).
Accompanying backend PR: https://github.com/openedx/ecommerce/pull/4115 

**Main changes:**
- Adding new error handling for Dynamic Payment Methods, where the Payment Intent status can get into unexpected state.
- New error message in case the learner's location is in a buy now pay later (BNPL) supported country but their billing address is in a country that BNPL does not support.
- Make postal code required for BNPL countries.
- Add shipping address to the Payment Intent since it's required for Afterpay.
- Add the `PaymentMethodMessagingElement`
- Get country code from location cookie.
- Refactor where we initialize Stripe to Checkout's parent to be able to redirect to receipt page (need to call Stripe to retrieve Payment Intent).
- Refactor `PaymentPage` component to handle redirect from BNPL, and redirect to ecommerce's receipt page.


.env
```diff
@@ -21,6 +21,8 @@ REFRESH_ACCESS_TOKEN_ENDPOINT=null
 SEGMENT_KEY=null
 SITE_NAME=null
 USER_INFO_COOKIE_NAME=null
+USER_LOCATION_COOKIE_NAME=null
```
julianajlk, line 24: This cookie is present in stage and production (checked the cookies in payment.edx.org) even though I did not see this added for payment MFE in `edx-internal` ([here](https://github.com/search?q=repo%3Aedx%2Fedx-internal%20USER_LOCATION_COOKIE_NAME&type=code))

src/payment/payment-methods/stripe/service.js
```diff
@@ -56,17 +71,20 @@ export default async function checkout(
           purchased_for_organization: purchasedForOrganization,
         },
       },
+      // Shipping is required for processing Afterpay payments
+      shipping: shippingAddress,
     },
   });
 
-  if (result.error?.code === 'payment_intent_unexpected_state' && result.error?.type === 'invalid_request_error') {
+  if (result.error) {
```
christopappas, line 79: previously this was checking for the presence of `.error` , just want to double check that that's no longer required

src/payment/data/handleRequestError.js
```diff
@@ -95,6 +95,19 @@ export default function handleRequestError(error) {
     ]);
   }
 
+  // Country not DPM compatible
+  if (error.type === 'invalid_request_error' && (
```
christopappas, line 99: this error is more specific than the basket changed error. should this come before the basket changed error just in case the previous condition is met somehow unexpectedly?

christopappas (approved):
I'm not super familiar with this codebase, but I have a few questions regarding error handling/objects. Probably not blockers but things I wanted ask

Also, how do you feel about the test coverage presently? Do you feel that the previous tests that were updated cover the cases here we expect with DPM?

src/payment/data/handleRequestError.js
julianajlk: Good idea, will move

src/payment/PaymentPage.jsx
```diff
@@ -46,15 +58,74 @@ class PaymentPage extends React.Component {
...
+    });
+  };
+
+  retrievePaymentIntentInfo = async () => {
+    // Get Payment Intent to retrieve the payment status and order number associated with this DPM payment.
+    // If this is not a Stripe dynamic payment methods (BNPL), URL will not contain any params
+    // and should not retrieve the Payment Intent.
+    const searchParams = new URLSearchParams(global.location.search);
+    const clientSecretId = searchParams.get('payment_intent_client_secret');
+    if (clientSecretId) {
```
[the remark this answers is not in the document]
julianajlk: Not currently, Stripe only appends the URL params when coming back from another payment environment (in our case BNPL options), for what we have as active in our account.

If in the future 3DS is done, we may have to modify this. But that's a big if.

src/payment/payment-methods/stripe/service.js
julianajlk: Still checking for the presence of `result.error` here, I modified this because this calls `handleApiError()` which calls `handleRequestError()' and the type of error is handled there. 

So the original code was just being more specific to one type of error and now I want to be more broad, but at the end the same type of error will still be handled.

zubair-ce07 (approved):
LGTM 👍

PR #820, opened 2023-10-21
feat!: remove Transifex calls for OEP-58 | FC-0012

hopewise:
## Breaking change!

This change breaks the Jenkins transifex integration which has been deprecated in favor of the new GitHub Transifex App integration as part of [OEP-58](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0058-arch-translations-management.html#specification).

## Changes

 - Removes all direct use of `tx pull` and `tx push` commands from the micro-frontend in favor
of the [`atlas pull`](https://github.com/openedx/openedx-atlas/) command.
 - Remove source and language translations from the repositories, hence no `.json` files will be committed into the repos. 
 - `src/i18n/index.js` should export and empty array so the `make pull_translations` override it with the dynamic list of languages.
 - Remove the experimental `OPENEDX_ATLAS_PULL` flag to make `atlas pull` the default. 
 - Remove all Transifex related `Makefile` targets and other files.

Test results
------------

 - [ ] Verify that `make pull_translations` works as expected.

<details><summary>make pull_translations test results</summary>

```
# I've run the following commands:
$ make requirements
$ make pull_translations
$ git diff

# Output of the commmands:

make pull_translations
rm -rf src/i18n/messages
mkdir src/i18n/messages
cd src/i18n/messages \
      && atlas pull \
               translations/frontend-component-header/src/i18n/messages:frontend-component-header  \
               translations/frontend-component-footer/src/i18n/messages:frontend-component-footer \
               translations/paragon/src/i18n/messages:paragon \
               translations/frontend-app-payment/src/i18n/messages:frontend-app-payment
Pulling translation files
 - directory: translations/frontend-component-header/src/i18n/messages:frontend-component-header translations/frontend-component-footer/src/i18n/messages:frontend-component-footer translations/paragon/src/i18n/messages:paragon translations/frontend-app-payment/src/i18n/messages:frontend-app-payment
 - repository: openedx/openedx-translations
 - branch: main
 - filter: Not Specified
 - expand-glob: Not Specified
Creating a temporary Git repository to pull translations into "./translations_TEMP"...
^[Done.
Setting git sparse-checkout rules...
Done.
Pulling translation files from the repository...
remote: Enumerating objects: 40, done.
remote: Counting objects: 100% (40/40), done.
remote: Compressing objects: 100% (22/22), done.
remote: Total 40 (delta 29), reused 25 (delta 18), pack-reused 0
[...]

PR #859, opened 2024-03-18
fix: regenerate `package-lock.json` to update dependencies

brian-smith-tcril:


OmarIthawi:
@brian-smith-tcril I've included this pull request into #820. I think it's better to close this one.

